Issue #5, opened 2023-12-20
Create functionality in GitSyncProcessor struct to create, update, and delete K8S resources


juliev0:

shubhamdixit863:
will work on it

chandankumar4:
I'll work on this issue

juliev0:
Note that for creating new resources, we will use the optional Namespace specified in the `destinations`, if it's defined. If it's not defined, we can assume that the resource is either a cluster-level resource, or the Namespace should be in the resource definition itself. (If the user actually specified a Namespace in both the resource definition and the `destinations` for this cluster, I guess we can decide if that's an error or we just override it and issue a warning maybe.)

juliev0:
@shubhamdixit863 you'll basically be calling these functions. @chandankumar4 once you decide on the function signatures, provide that information to Shubham. Thanks!

juliev0:
@chandankumar4 One thing I'm thinking about is the "update" function. Should we be calling the K8S API "Update" function or the "Patch" function when a resource is changed? What I'm wondering is: what if our Resource (say a Pod) has some new Label that's been added by K8S itself, or what if its Status has changed in some way? - what happens if we do an Update based on the new spec in Git? Would it override those changes? It may be okay to override the Label or anything else from the Spec, as perhaps it would simply be re-applied, but wouldn't want to override anything from the Status. May need to do some experimentation.

juliev0:
Okay, I think I have the answer to my question above wrt "Status". For Status, if you simply call Update() on the Spec, I believe it _should_ only touch the Spec and not the Status, which is a subresource according to [this](https://book-v1.book.kubebuilder.io/basics/status_subresource). This is at least what I've found through updating the GitSync Resource itself.

juliev0:
One complexity, though, is the "ResourceVersion" on the resource. This article is an interesting read: https://alenkacz.medium.com/kubernetes-operators-best-practices-understanding-conflict-errors-d05353dff421

chandankumar4:
okay!! I'll check wrt status that using Update() should not touch the status part

juliev0:
So, I guess it really is up to the individual resource CRD definition to determine whether Status is a subresource or not (i.e. a line like [this](https://github.com/numaproj-labs/numaplane/blob/main/config/crd/bases/numaplane.numaproj.io.github.com.numaproj-labs_gitsyncs.yaml#L107)). 

[...]

PR #13, opened 2024-01-04
GitSync Processor

shubhamdixit863:


internal/shared/git/gitsync_processor.go
```diff
@@ -0,0 +1,127 @@
...
+		if err != nil {
+			log.Printf("error checking for updates in the github repo %s", err)
+			// send the updates over channel
+			ch <- Message{Updates: updates, Err: err}
+		}
+
+	}
+}
+
+func NewGitSyncProcessor(gitSync *v1.GitSync, k8client client.Client) (*GitSyncProcessor, error) {
```
chandankumar4, line 87: can you try to run it once in local, and see if able to clone any public repo?

internal/shared/git/gitsync_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	})
+	if err != nil && !errors.Is(err, git.NoErrAlreadyUpToDate) {
+		return false, err
+	}
+	// Get local and remote references
+	localRef, err := r.Head()
+	if err != nil {
+		return false, err
+	}
+	remoteRef, err := r.Reference(plumbing.NewRemoteReferenceName("origin", "master"), true)
```
chandankumar4, line 54: git has now changed the default branch from `master` to `main`

internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,127 @@
...
+
+}
+
+func (processor *GitSyncProcessor) Create(gitSync v1.GitSync) error {
+	// creating   a new k8 resource
+	return nil
+}
+
+func (processor *GitSyncProcessor) Update(gitSync v1.GitSync) error {
+	for _, ch := range processor.channels {
```
juliev0, line 108: The intention for the `Update()` method was to handle the case of the GitSync CRD having changed. This one is TBD, but it would actually either send a message to the channels to shutdown, create a new goroutine with new corresponding channel, or send an "update repository" message to a channel. The place where we will listen for the channel change will be in `WatchRepo()`.

internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	// start monitoring the repository
+	ticker := time.NewTicker(3 * time.Minute)
+	for range ticker.C {
+		updates, err := checkForUpdates(repo)
+		if err != nil {
+			log.Printf("error checking for updates in the github repo %s", err)
+			// send the updates over channel
+			ch <- Message{Updates: updates, Err: err}
```
juliev0, line 81: No, we're not sending the message over the channel. Here we just do the work of applying/deleting the resources.

Rather, we should be listening to the channel actually, for a message telling us to shut ourselves down (and depending on if we go with Option 1 [here](https://docs.google.com/document/d/1d1xVkVwCT4sGijiL0ymboEcLuKTqYcg8mHh_9RjwECA/edit) a message to "update repository".

juliev0 (commented):
Sorry, I think maybe my document wasn't clear about how the channels are being used. We can discuss.

internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	})
+	if err != nil && !errors.Is(err, git.NoErrAlreadyUpToDate) {
+		return false, err
+	}
+	// Get local and remote references
+	localRef, err := r.Head()
+	if err != nil {
+		return false, err
+	}
+	remoteRef, err := r.Reference(plumbing.NewRemoteReferenceName("origin", "master"), true)
```
juliev0, line 54: I saw @chandankumar4 made a comment about this earlier, but I suppose some repos will use "master" and some will use "main" and maybe some could theoretically have a different default branch. Is there a way to determine what the default branch is called?

internal/git/gitsync_processor.go
juliev0: this could always be sort of a "TODO" for now

internal/git/gitsync_processor.go
xdevxy: I think this should be the information provided as part of the GitSync CRD. I will create an issue for it.

internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return &GitSyncProcessor{
+		gitSync:  *gitSync,
+		k8Client: k8client,
+		channels: channels,
+	}, nil
+
+}
+
+func (processor *GitSyncProcessor) Create(gitSync v1.GitSync) error {
```
xdevxy, line 102: `Create` name is a bit misleading as it is not clear whether this creates resource or creates GitSyncProcessor. Let's name it to `CreateResource` explicitly.

internal/git/gitsync_processor.go
xdevxy: We can probably just leave TODO here to follow up as it is part of Seasoned milestone. https://github.com/numaproj-labs/numaplane/issues/15

xdevxy (approved):

internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,127 @@
...
+	// Compare the references
+	if localRef.Hash() != remoteRef.Hash() {
+		return true, nil
+	}
+	return false, nil
+}
+
+func WatchRepo(repoPath v1.RepositoryPath, ch chan Message) {
+	// clone the repository
+	repo, err := git.PlainClone(repoPath.Path, false, &git.CloneOptions{
```
xdevxy, line 67: Why do you pass in `repoPath.Path` here? shouldn't it be a local path? Moreover, I'm not sure if we want to persist the change. Maybe we can use the in memory cloning ATM. @juliev0 any thoughts?

juliev0: Just looking at the go-git library myself for the first time. I guess you're more the expert on this @xdevxy so feel free to advise. I guess it does take a local path - but perhaps that could be equal to the `repo.Path`? 

I didn't realize there was an in-memory cloning option until now - are you thinking that will be faster then? That seems good if so.

internal/git/gitsync_processor.go
juliev0: @chandankumar4 will be defining the signature for this as part of [this](https://github.com/numaproj-labs/numaplane/issues/5) issue, so I don't think we need to include it here. (It shouldn't take the `GitSync` parameter anyway.)

xdevxy:
Should we close this one as you have a new PR @shubhamdixit863? Thanks!

PR #21, opened 2024-01-10
Add Reconciler functionality

### LINKED ISSUE #7: Create Controller reconciliation functionality to create, update, and delete GitSyncProcessor objects

juliev0:
Fixes #7 

Reconciler holds a map of GitSyncProcessors that it watches. A lock for each GitSync prevents any one from being modified concurrently by multiple goroutines calling `Reconcile()`. Furthermore, the map itself is a `sync.Map` since it will be touched by multiple goroutines.

Note that if the Controller were to die in the middle of an add/update, the K8S API will ensure that it gets reconciled again (I'm not sure if it would be immediate, or after a 20 minute delay (apparently, reconciliation happens automatically for every object every 20 minutes)).

If the Controller were to die in the middle of a deletion, since there is a Finalizer and since it only gets removed at the end of the deletion, then if it were to die prior to the finalizer being removed, another reconciliation would happen either immediately or after a 20 minute delay to ensure that we shutdown the GitSyncProcessor.

**Future work:**
- need to add validation for GitSync spec
- can look into reducing reconciliations when not needed


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: controller
+  newTag: latest
```
juliev0, line 8: not sure why this got modified

internal/controller/gitsync_controller.go
```diff
@@ -31,6 +39,32 @@ import (
...
+	// clusterName is the name of the cluster we're running on - used to determine which GitSyncs to store in memory
+	clusterName string
+}
+
+const (
+	finalizerName = "numaplane-controller"
+)
+
+func NewGitSyncReconciler(c client.Client, s *runtime.Scheme) (*GitSyncReconciler, error) {
+	clusterName, found := os.LookupEnv("CLUSTER_NAME") // TODO: if we incorporate a ConfigMap later, could include this in it
```
xdevxy, line 59: 👍

internal/controller/gitsync_controller.go
```diff
@@ -39,19 +73,190 @@ type GitSyncReconciler struct {
...
+
+	// We will have a GitSyncProcessor object for every GitSync that contains our cluster
+	// Therefore, we Create or Update it if our cluster is defined as a Destination in the GitSync
+	// We Delete it if either the GitSync is deleted or our cluster is removed from it
+
+	forThisCluster := gitSync.DeletionTimestamp.IsZero() && gitSync.Spec.ContainsClusterDestination(r.clusterName)
+	_, forThisClusterPrev := r.gitSyncProcessors.Load(gitSync.String())
+	shouldDelete := !forThisCluster && forThisClusterPrev
+	shouldAdd := forThisCluster && !forThisClusterPrev
+	shouldUpdate := forThisCluster && forThisClusterPrev
```
xdevxy, line 150: I think this check here also needs to contain namespace comparison?

internal/controller/gitsync_controller.go
```diff
@@ -39,19 +73,190 @@ type GitSyncReconciler struct {
...
+
+		processorAsInterface, _ := r.gitSyncProcessors.Load(gitSync.String())
+		processor := processorAsInterface.(*git.GitSyncProcessor)
+		logger.Infow("Updating existing GitSync", "GitSync", gitSync)
+		err = processor.Update(gitSync)
+		if err != nil {
+			logger.Errorw("Error updating GitSync", "err", err, "GitSync", gitSync)
+			gitSync.Status.MarkFailed("UpdateFailure", err.Error())
+			return ctrl.Result{}, err
+		}
```
xdevxy, line 203: Can we add a debug log if the update is successful?

internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+type GitSyncProcessor struct {
+}
+
+func NewGitSyncProcessor(gitSync *numaplanenumaprojiov1.GitSync, k8sClient client.Client) (*GitSyncProcessor, error) {
+	return &GitSyncProcessor{}, nil
+}
+
+func (processor *GitSyncProcessor) Update(gitSync *numaplanenumaprojiov1.GitSync) error {
```
xdevxy, line 16: Can you add a comment to Update and Delete to state that they need to be idempotent? Thanks!

xdevxy (commented):
Thanks a lot for doing it! Mostly good, just a few comments. :)

internal/controller/gitsync_controller_test.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+func init() {
+	_ = apiv1.AddToScheme(scheme.Scheme)
+	_ = appv1.AddToScheme(scheme.Scheme)
+	_ = corev1.AddToScheme(scheme.Scheme)
+}
+
+// test reconciliation of a new GitSync plus one that's been updated
+func Test_NewGitSync(t *testing.T) {
```
xdevxy, line 66: Add test cases to cover the update and delete as well.

internal/controller/gitsync_controller.go
```diff
@@ -39,19 +73,190 @@ type GitSyncReconciler struct {
...
+func (r *GitSyncReconciler) deleteGitSyncProcessor(ctx context.Context, gitSync *apiv1.GitSync) error {
+	logger := logging.FromContext(ctx)
+
+	if controllerutil.ContainsFinalizer(gitSync, finalizerName) {
+
+		// find it in the map and call Shutdown(), and then delete it from the map
+		processorAsInterface, found := r.gitSyncProcessors.Load(gitSync.String())
+		if !found {
+			logger.Warnw("Unexpected: GitSyncProcessor not found in map to delete it", "GitSync", gitSync)
+			return nil
```
whynowy, line 245: Remove the finalizer?

internal/controller/gitsync_controller.go
```diff
@@ -60,3 +265,21 @@ func (r *GitSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&apiv1.GitSync{}).
 		Complete(r)
 }
+
+// TODO: add validation
```
xdevxy, line 269: Can you create an issue for it? Thanks!

internal/controller/gitsync_controller.go
juliev0: How so? If we're listed as a cluster, we want a GitSyncProcessor set up. If we're not listed as a cluster, we don't. If the namespace were to change, that's something that the `GitSyncProcessor` will handle (making sure that the resources get transferred to the new namespace). Or at least that was my thought.

internal/controller/gitsync_controller.go
juliev0: just did!

api/v1/gitsync_types.go
```diff
@@ -97,3 +127,92 @@ type GitSyncList struct {
...
+			if reflect.DeepEqual(&condition, &c) {
+				return
+			}
+		}
+	}
+	condition.LastTransitionTime = metav1.NewTime(time.Now())
+	conditions = append(conditions, condition)
+	// Sort for easy read
+	sort.Slice(conditions, func(i, j int) bool { return conditions[i].Type < conditions[j].Type })
+	status.Conditions = conditions
```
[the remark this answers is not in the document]
juliev0: I copied this function directly from Numaflow. I'm not sure if I'm following what you're saying. If it's unclear, could I maybe just add some comments to it?

internal/controller/gitsync_controller.go
xdevxy: yeah, after a second thought, you're right. This is the checking for actions on GitSync CRs. And we only want GitSync CR to be cluster level thus namespace option should be ignored.  This actually leads to another question, that if somehow there are two GitSync CR with the same git repo path and destination in two separate namespaces. We should not create two GitSyncProcessors, right? Even one of them cannot apply the resources successfully. In that case we probably want to maintain a cluster level mapping between GitSync and GitSyncProcessor instead of namespaced one at L48.

juliev0: So, just to make sure we're on the same page, line 48 is basically maintaining the map of the namespaced name of the GitSync CRDs themselves. At this point, for our MVP our parent-level GitSync will indicate a single namespace (maybe "numaplane-system" itself), so all of the GitSyncs will basically end up in there. 

Here we're looking at the destinations described in the GitSync CRD itself, so these will be in other namespaces besides "numaplane-system" in general.

juliev0: So, you asked about if there are two GitSync CRs with the same git repo path and destination in two separate namespaces? Do you still have this question after my clarification?

juliev0: Actually, one thing I'm realizing I should do is have a good distinction in the Status for whether the CRD is applicable/not-applicable for this cluster.

xdevxy: > Actually, one thing I'm realizing I should do is have a good distinction in the Status for whether the CR is applicable/not-applicable for this cluster.

Yeah, that sounds good.

xdevxy: Thanks @juliev0 for the explanation! As we discussed offline. The thing I'm worried about is two GitSync CR unnecessarily watching the same git repo path and we definitely don't need to address it here. I will create an issue for it.

api/v1/gitsync_types.go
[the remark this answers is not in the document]
xdevxy: yeah, sure, comments are fine. This is more a nit. Thanks!

api/v1/gitsync_types.go
```diff
@@ -17,11 +17,28 @@ limitations under the License.
...
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+// +kubebuilder:validation:Enum="";Pending;Running;Failed
+type GitSyncPhase string
+
+type ConditionType string
+
+const (
+	GitSyncPhaseUnknown GitSyncPhase = ""
```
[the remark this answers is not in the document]
juliev0: yeah, maybe we don't need it. It would only be needed perhaps if we added a check to see if it was unset in the future.

internal/git/gitsync_processor.go
juliev0: I started doing that, but then I actually couldn't imagine an implementation for them that wouldn't be idempotent. Can you?

xdevxy: No, I don't have any concrete in mind. My intention was to ensure whoever pick this up have that concept in mind during implementation. But feel free to ignore it. We can always add it later.

config/manager/kustomization.yaml
juliev0, line 8: this seems to be getting generated by "make install" - not sure why it got changed but may not matter?

xdevxy (approved):

PR #26, opened 2024-01-11
GitsyncProcessor With in memory repository cloning

shubhamdixit863:


internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error cloning the repository url %s", err)
+	}
+	return r, nil
+}
+
+func NewGitSyncProcessor(gitSync *v1.GitSync, k8client client.Client, clusterName string) (*GitSyncProcessor, error) {
+	channels := make(map[string]chan Message)
+	for _, repo := range gitSync.Spec.RepositoryPaths {
+		gitCh := make(chan Message)
```
juliev0, line 59: I'm thinking maybe we should add a length to the channel (maybe 5?) or otherwise we'll block if we try to send something and there's no space available. If we were just going to send a message to shutdown on here, it might be okay, but I think we will most likely also include a message to update path once we do Option 1 [here](https://docs.google.com/document/d/1d1xVkVwCT4sGijiL0ymboEcLuKTqYcg8mHh_9RjwECA/edit#heading=h.uaz74auunyxn).

internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,72 @@
...
+		channels[repo.Name] = gitCh
+		_, err := cloneRepository(repo.RepoUrl, repo.TargetRevision)
+		if err != nil {
+			return nil, err
+		}
```
juliev0, line 64: so, is the plan to create the goroutine after cloning?

internal/git/gitsync_processor.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// reference can  can be a branch, a tag, or a commit hash
+func cloneRepository(repoUrl string, reference string) (*git.Repository, error) {
+	var referenceName plumbing.ReferenceName
+	if isValidTag(reference) {
+		referenceName = plumbing.ReferenceName(fmt.Sprintf("refs/tags/%s", reference))
+	} else if isCommitHash(reference) {
+		referenceName = plumbing.NewHashReference("", plumbing.NewHash(reference)).Name()
+	}
```
juliev0, line 44: we don't need to set "referenceName" in the case that we don't match either of those first two cases?

juliev0: also, should we allow the case of a commit hash? That should never change, right? And I'm not sure about the case of a tag? Should tags ever change? 

In general, it seems like we want to be monitoring a branch.

juliev0: I guess it seems like ArgoCD does do {commit, tag, branch} based on [this](https://argo-cd.readthedocs.io/en/stable/operator-manual/architecture/#repository-server). But I'm trying to wrap my head around the motivation for monitoring a commit hash or a tag. I'm wondering if we need to continue monitoring in that case or just clone and be done.

juliev0: Per [slack](https://intuit-teams.slack.com/archives/C05K82P3N9J/p1704998292080539) discussion, let's allow branch and tag, not commit.

internal/git/gitsync_processor.go
xdevxy: Right, shouldn't we clone within the goroutine?

juliev0: It seems like it would probably be okay to either clone within it or before starting it.

internal/git/gitsync_processor.go
shubhamdixit863: I have removed cloning as per commit hash ,now cloning on the basis of tag and branch @xdevxy @juliev0

juliev0: oh, I think @xdevxy (Hao) is saying above that she does want to allow for cloning a commit hash as well - it's just that in that case the goroutine wouldn't need to continue monitoring anything.

internal/git/gitsync_processor.go
```diff
@@ -1,19 +1,76 @@
...
+		SingleBranch:  true,
+		ReferenceName: plumbing.ReferenceName(reference),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error cloning the repository url %s", err)
+	}
+	return r, nil
+}
+
+func watchRepo(repoUrl string, reference string) {
```
juliev0, line 50: maybe for this method you can take the entire RepositoryPath struct, since you're going to need pretty much all of the fields from it?

shubhamdixit863: ok acknowledged

internal/git/gitsync_processor.go
```diff
@@ -1,19 +1,76 @@
...
+}
+
+func isValidTag(tag string) bool {
+	match, _ := regexp.MatchString(`^[vV]?\d+\.\d+\.\d+$`, tag)
+	return match
 }
 
-func NewGitSyncProcessor(gitSync *numaplanenumaprojiov1.GitSync, k8sClient client.Client) (*GitSyncProcessor, error) {
-	return &GitSyncProcessor{}, nil
+func cloneRepository(repoUrl string, reference string) (*git.Repository, error) {
```
juliev0, line 35: can you add a comment for the fact that we support tag, branch, and commit hash? (see my other comment below regarding commit hash)

internal/git/gitsync_processor.go
shubhamdixit863: So if the user is cloning on commit hash we dnt want to monitor it ? or how it would be

internal/git/gitsync_processor.go
```diff
@@ -1,19 +1,76 @@
...
 type GitSyncProcessor struct {
+	gitSync     v1.GitSync
+	channels    map[string]chan Message
+	k8Client    client.Client
+	clusterName string
+}
+
+func isValidTag(tag string) bool {
+	match, _ := regexp.MatchString(`^[vV]?\d+\.\d+\.\d+$`, tag)
+	return match
```
juliev0, line 32: Actually, do we need to impose these constraints on tags? Do we not allow a Tag which doesn't match this convention? (if we decide not to, perhaps we should add another field to the RepositoryPath to describe which one it is)

internal/git/gitsync_processor.go
juliev0: right, if it's a branch or a tag, we will continually check if it changed. But if it's a commit, it should never change, so there's no point to continually check. (At the moment, you haven't added any code here to continually check anyway of course.)

internal/git/gitsync_processor.go
shubhamdixit863: the problem would be the go git package ,it takes the reference as input and that can be the branch ,tag or commit hash so we need some kind of criteria to distinguish whether its a branch ,commit hash or tag @juliev0

juliev0: replied in [slack](https://intuit-teams.slack.com/archives/C05K82P3N9J/p1705422186374989)

internal/git/gitsync_processor.go
```diff
@@ -1,19 +1,128 @@
...
+	switch revision {
+	case "branch":
+		// monitor the branch
+	case "tag":
+	// monitor the tag for any change in commit
+
+	case "commit-hash":
+		// no monitoring
+
+	}
```
xdevxy, line 105: I think all of these can be change to the following(pseudo code)
```
	// In memory cloning
	r, err := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
		URL: ...,
	})
	if err != nil {
		// ...
	}

	h, err := r.ResolveRevision(plumbing.Revision(revision))
	if err != nil {
		// ...
	}

	w, err := r.Worktree()
	if err != nil {
		// ...
	}

	// ... checking out to the commit
	err = w.Checkout(&git.CheckoutOptions{
		Hash: *h,
	})
	if err != nil {
		// ...
	}
        // Apply to the change.
	if IsCommitSHA(revision) {
		// no monitoring
	} else {
                // monitor with intervals
        }
```

And in the Monitoring you can pull the latest change without cloning again.

shubhamdixit863: @juliev0 @xdevxy accomodated the changes suggested

internal/git/gitsync_processor.go
```diff
@@ -1,19 +1,84 @@
...
+	"log"
+	"regexp"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/storage/memory"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	numaplanenumaprojiov1 "github.com/numaproj-labs/numaplane/api/v1"
+	v1 "github.com/numaproj-labs/numaplane/api/v1"
```
juliev0, line 12: you probably don't need an alias in this case, since it's the same

internal/git/gitsync_processor.go
```diff
@@ -1,19 +1,84 @@
...
+		log.Println("monitoring")
+	}
+}
+
+func NewGitSyncProcessor(gitSync *v1.GitSync, k8client client.Client, clusterName string) (*GitSyncProcessor, error) {
+	channels := make(map[string]chan Message)
+	for _, repo := range gitSync.Spec.RepositoryPaths {
+		gitCh := make(chan Message, messageChanLength)
+		channels[repo.Name] = gitCh
+		go checkRevision(&repo)
```
juliev0, line 70: probably the goroutine method should be something like `watchRepo()` which could potentially invoke your `checkRevision()` method (although, I'm thinking `cloneRepo()` may be a better name for that)

shubhamdixit863: ok addressed @juliev0

internal/git/gitsync_processor.go
shubhamdixit863: there is some linting issue related to go imports if i am removing v1 .lint check fails -`File is not `goimports`-ed `(goimports) ,probably because imports tin generated files are not sorted that i also tried fixing but dint work,i will try to address this once its merged with main.

internal/git/gitsync_processor.go
```diff
@@ -1,19 +1,93 @@
...
+		Hash: *hash,
+	})
+	if err != nil {
+		return err
+	}
+
+	if isCommitSHA(revision) {
+		log.Println("No monitoring")
+	} else {
+		log.Println("monitoring")
```
juliev0, line 69: I assume these are just temporary print statements - otherwise, I'd say to make them `log.Debugf` or log.Debugw` and have more complete messages. But this is fine as a temporary thing.

internal/git/gitsync_processor.go
```diff
@@ -1,19 +1,93 @@
...
+	}
+	// TargetRevision can be a branch, a tag, or a commit hash
+	err = checkRevision(r, repoPath.TargetRevision)
+	if err != nil {
+		log.Fatalf("error  checking revision for the repository %s", err.Error())
+
+	}
+}
+
+func checkRevision(r *git.Repository, revision string) error {
```
juliev0, line 49: ultimately, this should probably also return a bool or something to indicate if we need to continually monitor it, but okay if you want to add once you implement the continual monitoring

internal/git/gitsync_processor.go
juliev0: okay. no big deal if you're not able to fix it

internal/git/gitsync_processor.go
shubhamdixit863: ok acknowledged

juliev0 (approved):

PR #28, opened 2024-01-11
chore: add github ci

### LINKED ISSUE #20: Add CI
Github Actions Workflow similar to [this](https://github.com/numaproj/numaflow/blob/main/.github/workflows/ci.yaml)

xdevxy:
Closes #20 

Adds CI with unit-test and lint as a starter.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,51 @@
+name: test
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+    branches:
+      - "main"
+      - "ci-test"
+jobs:
+  unit-tests:
+    name: Unit Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v4.0.0
+        with:
+          go-version: '1.20'
+        id: go
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+      - name: Restore Go build cache
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/go-build
+          key: ${{ runner.os }}-go-build-v1-${{ github.run_id }}
+      - name: Get dependencies
+        run: go mod download
+      - name: Test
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,51 @@
+name: test
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+    branches:
+      - "main"
+      - "ci-test"
```
xdevxy, line 9: Put it there to have a branch for any ci testing.

juliev0: that's a neat idea - so if we change this file in the future, we should use this same branch to do it?

juliev0: Sorry, what's the case where we would want to merge into "ci-test" instead of "main"?

xdevxy: this is only to ensure the CI change got triggered on this branch to verify it is correct. If we don't want it there, I can do a follow up to remove it.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,51 @@
...
+  unit-tests:
+    name: Unit Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v4.0.0
+        with:
+          go-version: '1.20'
+        id: go
```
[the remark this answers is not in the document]
xdevxy: :+1  thanks for the suggestion. I see this is a nice optimization. If you don't see strong reason to not do it manually.  Let's do it in a follow up.

xdevxy:
I think we should merge this as soon as possible, as it is already catching errors. See the lint failure.

.github/workflows/ci.yaml
xdevxy: Sorry @juliev0  I confused you! :D I misused this setting which is meant for identifying branch `targeting` to not currently on. Thanks for pointing out.

juliev0 (approved):

PR #36, opened 2024-01-16
feat: Add Kubernetes client for applying resource

chandankumar4:
Fixes: https://github.com/numaproj-labs/numaplane/issues/5


internal/kubernetes/client.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			return err
+		}
+		if err := resourceInfo.Refresh(obj, true); err != nil {
+			c.log.Warn(err)
+		}
+	}
+
+	// Patch an OpenAPI resource
+	_, patchedObj, err := patcher.Patch(resourceInfo.Object, modified, resourceInfo.Source, resourceInfo.Namespace,
```
chandankumar4, line 89: need to run the patch in case of resource is not available and it's just created because patch will update the last applied annotation, like below:
```bash
   annotations:
    deployment.kubernetes.io/revision: "1"
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"apps/v1","kind":"Deployment","metadata":{"annotations":{},"labels":{"app":"nginx","test":"nginx"},"name":"nginx","namespace":"numaplane-system"},"spec":{"replicas":1,"selector":{"matchLabels":{"app":"nginx"}},"template":{"metadata":{"labels":{"app":"nginx"}},"spec":{"containers":[{"image":"nginx","imagePullPolicy":"Always","name":"nginx"}]}}}}
```

PR #37, opened 2024-01-18
added some more fields to CommitStatus struct

juliev0:


api/v1/gitsync_types.go
```diff
@@ -92,10 +94,22 @@ type Destination struct {
...
+	// Hash of the git commit
 	Hash string `json:"hash"`
 
+	// Synced indicates if the sync went through
 	Synced bool `json:"synced,omitempty"`
+
+	// SyncTime represents the last time that we attempted to sync this commit (whether or not it succeeded)
+	SyncTime metav1.Time `json:"syncTime"`
+
+	// CommitTime is the time of the Git commit
```
xdevxy, line 108: What is the motivation to have this? I don't see how we will use it.

juliev0: if for some reason there were a delay in the system between CommitTime and SyncTime, it would reveal the delay. I think it's helpful, even though yes of course somebody could find it in the original Git commit, I think it's nice to see it here.

xdevxy: I feel it actually can be misleading, if the intention is to check delay. Because the commit time between commits doesn't necessarily go in order. One more recent commit can have a time earlier than an older one. https://stackoverflow.com/questions/63126817/does-git-ensure-commit-timestamp-order

xdevxy (approved):

Issue #25, opened 2024-01-11
Stores the recent commit in the `GitSync` Status


xdevxy:
After checkout a more recent commit and apply the resources according, we should store the commit hash and synced in the Status.

PR #43, opened 2024-01-19
GitSyncProcessor Iterate through all files with the given repo path

### LINKED ISSUE #4: Create goroutine which starts up and clones a git repository
Depends on #3 

Create a goroutine in GitSyncProcessor which will be used to sync a git repository. It will take a repository name and path, and an optional previous git commit hash. For now, just start up and clone the path. We assume whatever the main branch is.

If the previous git commit hash is passed in, clone the repository at that hash. If not, clone it at the latest HEAD.

Call the create and update functions described [here](https://github.com/numaproj-labs/numaplane/issues/5) depending on whether the resource already exists or not (if that work hasn't been done yet, just create no-op methods for those).

xdevxy:
Closes #4 

This PR iterates through all files with the given repo path so that the defined resources can be applied.


Dockerfile
```diff
@@ -24,7 +24,7 @@ ENV GOCACHE=/root/.cache/go-build
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
+RUN --mount=type=cache,target="/root/.cache/go-build" CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -o manager cmd/main.go
```
juliev0, line 27: Do you happen to know why they had the "-a" in there before? Is there some notion that you could somehow have outdated files or something?

juliev0: Is this a best practice for Dockerfiles?

internal/git/gitsync_processor.go
```diff
@@ -19,9 +21,11 @@ type Message struct {
 	Err     error
 }
 
-func isCommitSHA(revision string) bool {
-	match, _ := regexp.MatchString("^[0-9a-fA-F]{40}$", revision)
-	return match
+var commitSHARegex = regexp.MustCompile("^[0-9A-Fa-f]{40}$")
```
juliev0, line 24: good call

internal/git/gitsync_processor.go
```diff
@@ -31,60 +35,98 @@ type GitSyncProcessor struct {
...
+		return err
 	}
-	// TargetRevision can be a branch, a tag, or a commit hash
-	err = checkRevision(r, repoPath.TargetRevision)
+
+	// The revision can be a branch, a tag, or a commit hash
+	h, err := r.ResolveRevision(plumbing.Revision(repo.TargetRevision))
 	if err != nil {
-		log.Fatalf("error  checking revision for the repository %s", err.Error())
+		logger.Errorw("error resolve the revision", "revision", repo.TargetRevision, "err", err)
```
juliev0, line 53: "resolve" -> "resolving"

internal/git/gitsync_processor.go
```diff
@@ -31,60 +35,98 @@ type GitSyncProcessor struct {
...
 	})
 	if err != nil {
 		return err
 	}
 
-	if isCommitSHA(revision) {
-		log.Println("No monitoring")
+	if isCommitSHA(repo.TargetRevision) {
+		// TODO: no monitoring
+		logger.Debug("no monitoring")
```
juliev0, line 96: if we're keeping these log lines, can we add more information to them?

internal/git/gitsync_processor.go
```diff
@@ -31,60 +35,98 @@ type GitSyncProcessor struct {
...
+	}
+
+	for _, repo := range gitSync.Spec.RepositoryPaths {
+		gitCh := make(chan Message, messageChanLength)
+		channels[repo.Name] = gitCh
+		go func(repo *v1.RepositoryPath) {
+			err := watchRepo(ctx, repo, namespace)
+			if err != nil {
+				// TODO: Retry on non-fatal errors
+				logger.Errorw("error watch the repo", "err", err)
```
juliev0, line 124: "watch" -> "watching"

internal/git/gitsync_processor.go
xdevxy: I don't think we will keep the log, or at least we will know what to log in next iterations when we actually add the monitoring. The log here is only to avoid lint error.

xdevxy:
> Technically, this PR doesn't do this part of the original issue: "If the previous git commit hash is passed in, clone the repository at that hash. If not, clone it at the latest HEAD." so maybe we can add that as a separate issue?

Yeah, I will create a separate issue for it as it is relates to https://github.com/numaproj-labs/numaplane/issues/25

Dockerfile
xdevxy: > go help build
>   -a force rebuilding of packages that are already up-to-date.

I don't know why they had it in the first place. But looking at the definition of `-a`, it doesn't quite make sense to me to have to do it in this way.

juliev0 (approved):

PR #44, opened 2024-01-19
feat: Add Kubernetes client for applying resources

chandankumar4:
- Add kubernetes client for create/update resource
- Add kubernetes client for delete resource

Fixed: https://github.com/numaproj-labs/numaplane/issues/5


internal/kubernetes/client.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	gv := gvk.GroupVersion()
+	c.config.GroupVersion = &gv
+
+	restClient, err := newRestClient(*c.config, gv)
+	if err != nil {
+		return err
+	}
```
juliev0, line 51: In the future, I'm thinking we should explore maybe having a reusable TCP client rather than needing to start up a new connection with each apply.

chandankumar4: sure, I'll add TODO there, need to find some approach as it's generating restClient based on groupVersion of manifest.

chandankumar4:
I'll add another commit in this PR to call the `ApplyResource(...)` method from `watchRepo()` which will apply the resource in cluster.

internal/kubernetes/deploy.go
```diff
@@ -0,0 +1,86 @@
...
+		log:           logger,
+	}, nil
+}
+
+// DeleteResource will delete the resource from cluster using kind, name and namespace.
+func (c *client) DeleteResource(kind, name, namespace string, do metav1.DeleteOptions) error {
+	return c.deleteResourceByKindAndNameAndNamespace(kind, name, namespace, do)
+}
+
+// ApplyResource converts tha manifest byte data in unstructured format
```
juliev0, line 50:
```suggestion
// ApplyResource converts the manifest byte data to unstructured format and then applies it to the cluster
```

internal/kubernetes/client.go
```diff
@@ -0,0 +1,172 @@
...
+func computeDiscoverCacheDir(parentDir, host string) string {
+	// strip the optional scheme from host if it's there:
+	schemelesHost := strings.Replace(strings.Replace(host, "https://", "", 1), "http://", "", 1)
+	// now do a simple collapse of non-AZ09 characters.  Collisions are possible but unlikely.  Even if we do collide the problem is short lived
+	safeHost := overlyCautiousIllegalFileCharacters.ReplaceAllString(schemelesHost, "_")
+	return filepath.Join(parentDir, safeHost)
+}
+
+func SetNamespaceIfScoped(namespaceOverride string, u *unstructured.Unstructured, helper *resource.Helper) error {
+	if helper.NamespaceScoped {
```
juliev0, line 163: nice, so this is able to look at the ResourceType to determine if it's a namespaced resource?

internal/kubernetes/patcher.go
```diff
@@ -0,0 +1,421 @@
...
+	"k8s.io/klog/v2"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/scheme"
+	"k8s.io/kubectl/pkg/util"
+	"k8s.io/kubectl/pkg/util/openapi"
+)
+
+// All this code copied from
+// https://github.com/kubernetes/kubectl/blob/v0.29.0/pkg/cmd/apply/patcher.go
```
juliev0, line 35: did you ever get a chance to try using their Patcher rather than copying? (not that I want to necessarily hold things up for that)

internal/kubernetes/client.go
chandankumar4: yes

internal/kubernetes/patcher.go
chandankumar4: yes, I tried it once but I got some error while testing out and didn't get time to look further, I'll check it back again after this PR, I'll create an issue to track it.

chandankumar4: https://github.com/numaproj-labs/numaplane/issues/49

juliev0: Thanks!

internal/kubernetes/client.go
xdevxy: Cool!

internal/git/gitsync_processor.go
```diff
@@ -123,14 +130,13 @@ func watchRepo(ctx context.Context, repo *v1.RepositoryPath, _ /* namespace */ s
 	return nil
 }
 
-func NewGitSyncProcessor(ctx context.Context, gitSync *v1.GitSync, k8client client.Client, clusterName string) (*GitSyncProcessor, error) {
+func NewGitSyncProcessor(ctx context.Context, gitSync *v1.GitSync, config *rest.Config, clusterName string) (*GitSyncProcessor, error) {
```
juliev0, line 133: I'm not sure, but I wonder if we still want to have the `k8client client.Client` in there or not? (which is I guess for `GitSyncs` in particular) Or alternatively the Status subresource used [here](https://github.com/numaproj-labs/numaplane/blob/main/internal/controller/gitsync_controller.go#L127)? We also need to update the `Status.CommitStatus`. Or did you imagine we would derive a client from the config for that as well?

internal/git/gitsync_processor.go
```diff
@@ -99,14 +106,14 @@ func watchRepo(ctx context.Context, repo *v1.RepositoryPath, _ /* namespace */ s
 	// Read all the files under the path and apply each one respectively.
 	err = tree.Files().ForEach(func(f *object.File) error {
 		logger.Debugw("read file", "file_name", f.Name)
-		_, err = f.Contents()
+		manifest, err := f.Contents()
```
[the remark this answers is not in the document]
juliev0: (@shubhamdixit863 is doing this)

internal/kubernetes/client.go
juliev0: thanks, I just wrote this issue: https://github.com/numaproj-labs/numaplane/issues/52

internal/git/gitsync_processor.go
chandankumar4: I'm thinking to have one client which we can derive from rest config but I'm open for both as well, if we want to keep k8sclient also I'll add that?

juliev0: I see it's nice to eliminate redundancy - but it does also save some steps not to have to derive the client. 
One thing is: I am now thinking that `k8client client.Client` and `config *rest.Config` could use more meaningful names. 

I think `k8client` may only be for numaplane CRDs? But I haven't tested to verify it. If so, maybe we should have a name that reflects that?

And then `config` could use a more meaningful name as well. 

What do you think? 

Also, by the way, if you do prefer not to have the redundancy of the two parameters, can you create the client for the subresource so it can be used by @xdevxy 's PR?

chandankumar4: I'm thinking to refactor it a bit like from main file itself I'll generate this kubernetes.client and pass it all over where ever it required and so that I don't have to pass restConfig as well as k8sClient all over and also I have a method inside client interface defined like Get, Delete and others? probably we can discuss more over call?

For now I'm thinking to not remove the k8sClient from here and keep both because refactor PR will be a bit longer?

internal/kubernetes/deploy_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/numaproj-labs/numaplane/tests/utils"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestToUnstructured(t *testing.T) {
```
xdevxy, line 11: Can you add test cases for Apply/Delete resources as well? Is it possible to do it with the fake k8s client?

xdevxy: This can be a follow up though.

xdevxy (approved):
Overall looks good, all comments can be a followup.

internal/kubernetes/deploy_test.go
chandankumar4: yes, I'll have a follow up PR for fake k8s client as I have added TODO here https://github.com/numaproj-labs/numaplane/pull/44/files#diff-321de43a0cf1c625281d5f154b8fb166a5f4984578474d9bf0758dd4a2aab75dR18

PR #50, opened 2024-01-22
chore: Fetch all remote branches for resolution 

xdevxy:
This PR adds the logic to fetch all remote branches for correct `TargetRevision` resolution. It also allows root directory as repo `Path`.

### DIFF
--- a/internal/git/gitsync_processor_test.go
+++ b/internal/git/gitsync_processor_test.go
@@ -15,7 +15,7 @@ func Test_watchRepo(t *testing.T) {
 		hasError bool
 	}{
 		{
-			name: "branch name as a TargetRevision",
+			name: "`main` as a TargetRevision",
 			repo: v1.RepositoryPath{
 				RepoUrl:        "https://github.com/numaproj-labs/numaplane.git",
 				Path:           "config/samples",
@@ -45,16 +45,16 @@ func Test_watchRepo(t *testing.T) {
 			name: "remote branch name as a TargetRevision",
 			repo: v1.RepositoryPath{
 				RepoUrl:        "https://github.com/git-fixtures/basic.git",
-				Path:           "./",
+				Path:           "go",
 				TargetRevision: "refs/remotes/origin/branch",
 			},
 			hasError: false,
 		},
 		{
-			name: "more branch name as a TargetRevision",
+			name: "local branch name as a TargetRevision",
 			repo: v1.RepositoryPath{
 				RepoUrl:        "https://github.com/git-fixtures/basic.git",
-				Path:           ".",
+				Path:           "go",
 				TargetRevision: "branch",
 			},
 			hasError: false,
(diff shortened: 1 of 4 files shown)


internal/git/gitsync_processor_test.go
```diff
@@ -33,11 +42,29 @@ func Test_watchRepo(t *testing.T) {
...
-				Path:           ".github",
-				TargetRevision: "v5.5.1",
+				RepoUrl:        "https://github.com/git-fixtures/basic.git",
+				Path:           "./",
+				TargetRevision: "refs/remotes/origin/branch",
+			},
+			hasError: false,
+		},
+		{
+			name: "more branch name as a TargetRevision",
```
juliev0, line 54:
```suggestion
			name: "local branch name as a TargetRevision",
```

juliev0: (wasn't sure what you meant)

juliev0:
I wonder what we should allow? All 3 of those? Should we also therefore allow "./something" as well as "/something"?

internal/git/gitsync_processor_test.go
xdevxy: Meant for more test cases for branch name as TargetRevision for resolution. But your suggestion is meaningful. Thanks!

xdevxy:
> I wonder what we should allow? All 3 of those? Should we also therefore allow "./something" as well as "/something"?

Yeah, makes sense. I was wondering the same question as well. Now I think about it, we probably just want to support `"" `as the current root dir to be consistent with how we parse the path. And to be clear, I will add a comment to the `Path` variable in the spec to clarify.

juliev0:
FYI, I mentioned in Shubham's PR [here](https://github.com/numaproj-labs/numaplane/pull/47/files#r1462237919) that it would be great if these log lines could maybe include the GitSync name and RepositoryPath name in them.

juliev0:
> > I wonder what we should allow? All 3 of those? Should we also therefore allow "./something" as well as "/something"?
> 
> Yeah, makes sense. I was wondering the same question as well. Now I think about it, we probably just want to support `"" `as the current root dir to be consistent with how we parse the path. And to be clear, I will add a comment to the `Path` variable in the spec to clarify.

Yeah, I think that sounds good. (BTW, it was probably in some of my design documentation that I was also being inconsistent with this, since I didn't really know how it would be)

xdevxy:
> > > I wonder what we should allow? All 3 of those? Should we also therefore allow "./something" as well as "/something"?
> > 
> > 
> > Yeah, makes sense. I was wondering the same question as well. Now I think about it, we probably just want to support `"" `as the current root dir to be consistent with how we parse the path. And to be clear, I will add a comment to the `Path` variable in the spec to clarify.
> 
> Yeah, I think that sounds good. (BTW, it was probably in some of my design documentation that I was also being inconsistent with this, since I didn't really know how it would be)

Which design doc you are referring to? In general, I feel these are too detailed and fine with left it out from design doc. :)

xdevxy:
> FYI, I mentioned in Shubham's PR [here](https://github.com/numaproj-labs/numaplane/pull/47/files#r1462237919) that it would be great if these log lines could maybe include the GitSync name and RepositoryPath name in them.

Let's address this in follow ups. I will refactor the code to support the other issues. I will keep in mind the log at that time.

juliev0 (approved):

Issue #45, opened 2024-01-19
NewGitSyncProcessor from last persisted commit status


xdevxy:
When start to watch a repo in GitSyncProcessor, if a git commit status is persisted in the GitSync CR, clone the repository at that hash and check if any diff happens since then. If not, clone it at the TargetRevision (existing logic).

xdevxy:
Close with https://github.com/numaproj-labs/numaplane/pull/51

Issue #46, opened 2024-01-19
Worker goroutines to reduce the number of concurrent goroutines reading from RepositoryPaths


juliev0:
This would take care of the following issues:

- be sure to avoid the case where a given goroutine monitoring a RepositoryPath returns silently
- reduce the number of goroutines starting up all at once at the beginning and cloning repos all at once

juliev0:
One way to do it:
Could still have a map of GitSyncs to store state (maybe a map of maps where the inner map is the RepoPath or a map of GitSyncProcessors which contains a map of RepoPaths). The reconcile functionality would just update the state of memory in that map, plus if there are any new RepositoryPaths (whether they come from a new GitSyncs or an existing GitSync), add them to a channel. The workers could look up the state in the map and if the RepoPath or GitSync has been deleted, just return.

For ensuring that we don't pull from github more than once every 3 minutes, we could store the last time in the state, and return in the case that not enough time has gone by.

To ensure synchronous processing within a repository path, the goroutine could add itself back to the channel at the end.

PR #51, opened 2024-01-23
feat: Stores the recent commit in the GitSync Status

### LINKED ISSUE #25: Stores the recent commit in the `GitSync` Status

xdevxy:
Closes #25 and #45 

This PR stores the recent commit if the resources successfully applied. If not, the whole process will be retried.

It also ensures with a previous persisted commit status, we just monitor and apply the changes in-between.

### DIFF
--- a/internal/git/gitsync_processor.go
+++ b/internal/git/gitsync_processor.go
@@ -3,12 +3,16 @@ package git
 import (
 	"context"
 	"regexp"
+	"time"
 
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/go-git/go-git/v5/storage/memory"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -44,22 +48,20 @@ type GitSyncProcessor struct {
 	clusterName string
 }
 
-func watchRepo(ctx context.Context, restConfig *rest.Config, repo *v1.RepositoryPath, namespace string) error {
-	logger := logging.FromContext(ctx)
+func cloneRepo(repo *v1.RepositoryPath) (*git.Repository, error) {
+	return git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
+		URL: repo.RepoUrl,
+	})
+}
+
+func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, restConfig *rest.Config, k8Client client.Client, repo *v1.RepositoryPath, namespace string) error {
+	logger := logging.FromContext(ctx).With("GitSync name", gitSync.Name,
+		"RepositoryPath Name", repo.Name)
 
 	// create kubernetes client
-	k8sClient, err := kubernetes.NewClient(restConfig, logger)
+	client, err := kubernetes.NewClient(restConfig, logger)
 	if err != nil {
 		logger.Errorw("cannot create kubernetes client", "err", err)
-		return err
-	}
-
-	r, err := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
-		URL:          repo.RepoUrl,
-		SingleBranch: false,
-	})
-	if err != nil {
-		logger.Errorw("error cloning the repository", "err", err)
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


internal/git/gitsync_processor.go
```diff
@@ -42,17 +44,14 @@ type GitSyncProcessor struct {
...
+		URL: repo.RepoUrl,
 	})
-	if err != nil {
-		logger.Errorw("error cloning the repository", "err", err)
-		return err
-	}
+}
+
+func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, k8Client client.Client, repo *v1.RepositoryPath, _ /* namespace */ string) error {
+	logger := logging.FromContext(ctx)
```
juliev0, line 54: can we set this logger to include GitSync name and RepositoryPath name in it using `logging.FromContext(ctx).With(...)`? Then we won't need to add it to each logger line below

internal/git/gitsync_processor.go
```diff
@@ -139,10 +162,23 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1.GitSync, k8client clie
...
-				// TODO: Retry on non-fatal errors
-				logger.Errorw("error watching the repo", "err", err)
+				logger.Errorw("error cloning the repo",
+					"GitSync name", gitSync.Name,
+					"RepositoryPath Name", repo.Name,
+					"err", err)
+			} else {
+				for ok := true; ok; {
+					err = watchRepo(ctx, r, gitSync, k8client, repo, namespace)
+					// TODO: Break on fatal errors
```
juliev0, line 174: If there were a fatal error in this goroutine (which I'm not sure if there would be?), wouldn't the inner code just directly call `log.Fatal()` and then the whole thing would shut down? Are you thinking we need to check for it here for some reason?

internal/git/gitsync_processor.go
```diff
@@ -67,55 +66,79 @@ func watchRepo(ctx context.Context, repo *v1.RepositoryPath, _ /* namespace */ s
...
+			Hash:     hash.String(),
+			Synced:   true,
+			SyncTime: metav1.NewTime(time.Now()),
+		}
+		if gitSync.Status.CommitStatus == nil {
+			gitSync.Status.CommitStatus = make(map[string]v1.CommitStatus)
+		}
+		gitSync.Status.CommitStatus[repo.Name] = commitStatus
+		// Errors will be retried
+		if err = k8Client.Update(ctx, gitSync); err != nil {
```
juliev0, line 127: So, we need to actually be updating the Status subresource like [this](https://github.com/numaproj-labs/numaplane/blob/main/internal/controller/gitsync_controller.go#L127)

Also, when you say "Errors will be retried" in your comment, are you saying that since we have the map, we will automatically retry again the next time we come through in 3 minutes?

internal/git/gitsync_processor.go
```diff
@@ -67,55 +66,79 @@ func watchRepo(ctx context.Context, repo *v1.RepositoryPath, _ /* namespace */ s
...
+		if err != nil {
+			return err
+		}
+		commitStatus := v1.CommitStatus{
+			Hash:     hash.String(),
+			Synced:   true,
+			SyncTime: metav1.NewTime(time.Now()),
+		}
+		if gitSync.Status.CommitStatus == nil {
+			gitSync.Status.CommitStatus = make(map[string]v1.CommitStatus)
```
juliev0, line 123: unfortunately, I think we may need a lock around this map since we have multiple goroutines touching it.

juliev0, line 123: It's not thread safe to update the central map. So, I guess there are 2 options for dealing with this:
1. Continue to update gitSync.Status.CommitStatus but in a thread safe way
2. Maintain a copy of the gitSync.Status here instead of a reference to the actual object. Each time we want to update GitSync.Status, we call out to do a Get() in order to get a local copy, modify it, and then do an update.

juliev0: Note that for the 2nd option, we don't need to be concerned about the extra API call - since there is a local informer cache that holds the value.

juliev0: I haven't thought about it thoroughly but thinking like we may need to go with the 2nd approach - if we do the 1st approach, we'd need to make sure there's no way that one goroutine's CommitStatus could fail to get updated.

internal/git/gitsync_processor.go
xdevxy: Yeah, meant to use `Status()` client but cannot get the test working. Will look into that a bit more.

internal/git/gitsync_processor.go
xdevxy: Yeah, I think the second is cleaner and the way to go. 👍

internal/git/gitsync_processor.go
xdevxy: Sorry I should make it more clear the fatal errors I put here is for the GitSyncProcessor goroutine, not the entire process should shut down in this case.

internal/git/gitsync_processor.go
```diff
@@ -76,56 +78,107 @@ func watchRepo(ctx context.Context, restConfig *rest.Config, repo *v1.Repository
...
-		// Apply manifest in cluster
-		return k8sClient.ApplyResource([]byte(manifest), namespace)
-	})
-	if err != nil {
-		return err
+		commitStatus := v1.CommitStatus{
+			Hash:     hash.String(),
+			Synced:   true,
+			SyncTime: metav1.NewTime(time.Now()),
+		}
```
juliev0, line 149: it looks like we won't log a CommitStatus with Synced=false if we fail?

internal/git/gitsync_processor.go
```diff
@@ -76,56 +78,107 @@ func watchRepo(ctx context.Context, restConfig *rest.Config, repo *v1.Repository
...
-		//TODO: this currently assumes that one file contains just one manifest - modify for multiple
-		manifest, err := f.Contents()
+		if !isRootDir(repo.Path) {
+			// Locate the tree with the given path.
+			tree, err = tree.Tree(repo.Path)
+			if err != nil {
+				logger.Errorw("error locate the path", "repository path", repo.Path, "err", err)
+				return err
+			}
+		}
```
juliev0, line 126: To make @shubhamdixit863 's job of merging a little easier, can you add this same `getCommitTreeAtPath()` method from his [PR](https://github.com/numaproj-labs/numaplane/pull/47/files) and call it here?

internal/git/gitsync_processor.go
```diff
@@ -76,56 +78,107 @@ func watchRepo(ctx context.Context, restConfig *rest.Config, repo *v1.Repository
...
+		gitSync = &v1.GitSync{}
+		if err = k8Client.Get(ctx, namespacedName, gitSync); err != nil {
+			// if we aren't able to do a Get, then either it's been deleted in the past, or something else went wrong
+			if apierrors.IsNotFound(err) {
+				return nil
+			} else {
+				logger.Errorw("Unable to get GitSync", "err", err)
+				return err
+			}
+		}
```
juliev0, line 160: why do we do the `Get()` again if we just did it? Maybe we should do the `Update()` and then if it fails, repeatedly do `Get()` and then `Update()` in an exponential backoff? And put the whole thing in a method so that @shubhamdixit863 will be able to call it to?

internal/git/gitsync_processor.go
```diff
@@ -76,56 +78,107 @@ func watchRepo(ctx context.Context, restConfig *rest.Config, repo *v1.Repository
 	}
 
 	// The revision can be a branch, a tag, or a commit hash
-	h, err := r.ResolveRevision(plumbing.Revision(repo.TargetRevision))
+	hash, err := r.ResolveRevision(plumbing.Revision(repo.TargetRevision))
```
juliev0, line 81: say somebody has a bad `TargetRevision` specified - we will be in a tight loop of retrying over and over 

This is why we should distinguish transient from non-transient errors now, and also employ exponential backoff. Let's make sure to add an Issue to do it if we don't do it now.

xdevxy: yeah, that is why I added the TODO, will create a ticket for it. :)

juliev0:
What's the plan for testing all of the various conditions? Do we need a follow up issue?

internal/git/gitsync_processor.go
xdevxy: yeah, sure, but would prefer to do it in a follow up PR. I can even do it after his PR is merged. That would be easier for all of us.

juliev0:
@shubhamdixit863 make sure you look at this PR so you have a heads up, if you haven't already. thanks!

internal/git/gitsync_processor.go
juliev0: I think if yours is merged, then he will need to do it, but maybe that's okay

internal/git/gitsync_processor.go
xdevxy: https://github.com/numaproj-labs/numaplane/issues/59

internal/git/gitsync_processor.go
xdevxy: yeah, good question, I feel that is not necessary, as no record can mean we   are not synced and we may get errors when we commit status with `synced=false` . But it is a nice to have though. HDYT?

internal/git/gitsync_processor.go
```diff
@@ -76,56 +78,107 @@ func watchRepo(ctx context.Context, restConfig *rest.Config, repo *v1.Repository
...
+		} else {
+			logger.Errorw("Unable to get GitSync", "err", err)
+			return err
+		}
 	}
+	val, ok := gitSync.Status.CommitStatus[repo.Name]
+	// Only create the resources for the first time if not created yet.
+	// Otherwise, monitoring with intervals.
+	if !ok {
+		// Retrieving the commit object matching the hash.
```
juliev0, line 105:
```suggestion
		// TODO: call getCommitTreeAtPath() here
		// Retrieving the commit object matching the hash.
```

internal/git/gitsync_processor.go
juliev0: It's true that if we have an error involving Kubernetes, then logging commit status probably won't work anyway. But if we have an error involving Git, then it would. Actually, there is an [Error](https://github.com/numaproj-labs/numaplane/blob/main/api/v1/gitsync_types.go#L110) field in the CommitStatus. We should be using that.

xdevxy: Ah that's a good point, that's the intention.

juliev0: Okay, please create an Issue

internal/git/gitsync_processor.go
xdevxy: Add a method for update commit status specifically. For exponential backoff, we discussed offline, it would be address in https://github.com/numaproj-labs/numaplane/issues/46 as we are switching to work queue model.

xdevxy:
> What's the plan for testing all of the various conditions? Do we need a follow up issue?

I think the test cases can be re-enable and add more after https://github.com/numaproj-labs/numaplane/issues/56

juliev0 (approved):
will approve since some of the issues will be addressed in follow up PRs

xdevxy:
> @shubhamdixit863 make sure you look at this PR so you have a heads up, if you haven't already. thanks!

@shubhamdixit863 yeah let me know if you need help with the rebase. Thanks!

Issue #30, opened 2024-01-11
K8S Error returned from Reconciler Status Update: "the object has been modified"


juliev0:
Sometimes Kubernetes returns an error when the Reconciler does an Update of Status: "the object has been modified". Luckily, a subsequent reconciliation always seems to occur and succeeds, but need to find out why this is happening.

juliev0:
I noticed this happened when I created a new one, which caused a Reconciliation in which I updated the Status (using the Status SubresourceWriter), and the resulting ResourceVersion was 3687148. 
Then another Reconcile got called, and when it called Get(), it got the previous version 3687147 instead of 3687148. When it tried to update the Status this time, it got the error. 

I wonder if the GitSync and the Status Subresource are stored in two different caches.

juliev0:
this looks to be someone who had the exact same issue: https://github.com/kubernetes-sigs/controller-runtime/issues/741

Most of these cases should stop if we set the resource's Watch to explicitly ignore `Status()` changes (meaning to only have `Reconcile()` called when there's a Generation change ([this](https://github.com/numaproj-labs/numaplane/issues/23) issue).

juliev0:
not sure why the PR didn't close this, so I'm closing manually

PR #54, opened 2024-01-24
Reduce reconciliations

### LINKED ISSUE #23: Avoid unnecessary reconciliations
The ISBService Controller performs a watch, ignoring when the Generation has not changed: https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/cmd/start.go#L103

This apparently causes changes to the Status subresource not to invoke `Reconcile()`, which will be useful if we're storing recent commit information in the Status.

Currently, we don't make an explicit `Watch()` call, so it must be happening under the hood.

juliev0:
Fixes #23 #30 

Explicitly watch just the _Generation Change_ on the resource, which will essentially only notice if the Spec has changed, not the Status.


juliev0:
tested changing Spec - seems like we now only call `Reconcile()` once when the Spec changes (as opposed to as a result of adding the Finalizer or as a result of updating the Status)

xdevxy (approved):

xdevxy:
> tested changing Spec - seems like we now only call `Reconcile()` once when the Spec changes (as opposed to as a result of adding the Finalizer or as a result of updating the Status)

It can be a follow up, do you think we can have a unit test for it? Thanks!

PR #66, opened 2024-01-29
Add env variable for CLUSTER_NAME

chandankumar4:
Deployed controller specs

```bash
macos-P9DJ3XY0QL :: github.com/numaproj-labs/numaplane ‹cluster-name-env*› » k get deploy numaplane-controller-manager -n numaplane-system -o yaml

      ......
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      - args:
        - --health-probe-bind-address=:8081
        - --metrics-bind-address=127.0.0.1:8080
        - --leader-elect
        command:
        - /manager
        env:
        - name: CLUSTER_NAME
          value: staging-usw2-k8s
        imagePullPolicy: Always
      .....
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 # Image URL to use all building/pushing image targets
 IMG ?= numaplane-controller
 VERSION ?= latest
-# Default cluster name where pumaplane get deployed, update it as needed.
+# Default cluster name where numaplane get deployed, update it as needed.
 CLUSTER_NAME ?= "staging-usw2-k8s"
 #TODO: add back once ready to push to quay.io
 #IMAGE_NAMESPACE ?= quay.io/numaproj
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,6 +2,8 @@
 # Image URL to use all building/pushing image targets
 IMG ?= numaplane-controller
 VERSION ?= latest
+# Default cluster name where pumaplane get deployed, update it as needed.
```
juliev0, line 5: fix spelling of "numaplane"

config/manager/manager.yaml
```diff
@@ -70,6 +70,9 @@ spec:
         - /manager
         args:
         - --leader-elect
+        env:
+          - name: CLUSTER_NAME
+            value: CLUSTER_NAME_VALUE
```
juliev0, line 75: maybe we should add a comment here to replace it with one's actual cluster name?

Makefile
chandankumar4: somehow spell check is not working in my editor, I need to fix it.

juliev0 (approved):

xdevxy (approved):

Issue #69, opened 2024-01-31
'object not found' error when watching a repo


xdevxy:
When watching a repo, periodically getting the following errors running in a local cluster,
```
{"level":"error","ts":"2024-01-31T00:53:47.905628359Z","logger":"numaplane","caller":"git/gitsync_processor.go:179","msg":"checking for new updates err","GitSync name":"gitsync-example","RepositoryPath Name":"user-repo","err":"object not found","stacktrace":"github.com/numaproj-labs/numaplane/internal/git.watchRepo\n\t/workspace/internal/git/gitsync_processor.go:179\ngithub.com/numaproj-labs/numaplane/internal/git.NewGitSyncProcessor.func1\n\t/workspace/internal/git/gitsync_processor.go:481"}
{"level":"error","ts":"2024-01-31T00:53:47.905639776Z","logger":"numaplane","caller":"git/gitsync_processor.go:484","msg":"error watching the repo","err":"object not found","stacktrace":"github.com/numaproj-labs/numaplane/internal/git.NewGitSyncProcessor.func1\n\t/workspace/internal/git/gitsync_processor.go:484"}
```

juliev0:
So, you think the object being referred to is the GitSync itself? On which line are we referring to it that would cause the problem?

juliev0:
It does seem problematic that we are referring to `status.CommitStatus` in `CheckForRepoUpdates()` if that's what you're referring to. I think we should just be saving the lastCommitStatus in the object so we don't need to do a `Get` to the API every 3 minutes.

juliev0:
You could actually just store the lastCommitHash locally on the stack too. Instead of calling:
```
CheckForRepoUpdates(ctx, r, repo, &gitSync.Status, namespace)
```
you could call:
```
CheckForRepoUpdates(ctx, r, repo, &lastCommitHash, namespace)
```

xdevxy:
yeah, exactly

> You could actually just store the lastCommitHash locally on the stack too. Instead of calling:
> 
> ```
> CheckForRepoUpdates(ctx, r, repo, &gitSync.Status, namespace)
> ```
> 
> you could call:
> 
> ```
> CheckForRepoUpdates(ctx, r, repo, &lastCommitHash, namespace)
> ```
> 
> Or..I do see that you are returning the hash currently, so if you want to continue doing that, then pass it by value:
> 
> ```
> CheckForRepoUpdates(ctx, r, repo, lastCommitHash, namespace)
> ```
[...]

PR #71, opened 2024-01-31
Object Not found Issue 69

shubhamdixit863:
https://github.com/numaproj-labs/numaplane/issues/69

### DIFF
--- a/internal/git/gitsync_processor.go
+++ b/internal/git/gitsync_processor.go
@@ -109,7 +109,7 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, rest
 	lastCommitHash = hash.String()
 	namespacedName := types.NamespacedName{
 		Namespace: gitSync.Namespace,
-		Name:      gitSync.Name,
+		Name:      gitSync.Name,/
 	}
 	gitSync = &v1.GitSync{}
 	if err = k8Client.Get(ctx, namespacedName, gitSync); err != nil {
@@ -172,17 +172,18 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, rest
 				if err != nil {
 					return err
 				}
-
+                lastCommitHash=recentHash
 				// recentHash would be an empty string if there is no update in the  repository
 				if len(recentHash) > 0 {
-					err = updateCommitStatus(ctx, k8Client, namespacedName, recentHash, repo, logger)
-					if err != nil {
-						return err
-					}
 					err = ApplyPatchToResources(patchedResources, client)
 					if err != nil {
 						return err
 					}
+					err = updateCommitStatus(ctx, k8Client, namespacedName, recentHash, repo, logger)
+					if err != nil {
+						return err
+					}
+
 
 				}
 			case <-ctx.Done():
--- a/internal/git/gitsync_processor.go
+++ b/internal/git/gitsync_processor.go
@@ -109,7 +109,7 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, rest
 	lastCommitHash = hash.String()
 	namespacedName := types.NamespacedName{
 		Namespace: gitSync.Namespace,
-		Name:      gitSync.Name,/
+		Name:      gitSync.Name,
 	}
 	gitSync = &v1.GitSync{}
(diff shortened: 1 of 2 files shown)


internal/git/gitsync_processor.go
```diff
@@ -168,17 +168,17 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, rest
...
-				err = updateCommitStatus(ctx, k8Client, namespacedName, recentHash, repo, logger)
-				if err != nil {
-					return err
-				}
 				// recentHash would be an empty string if there is no update in the  repository
 				if len(recentHash) > 0 {
+					err = updateCommitStatus(ctx, k8Client, namespacedName, recentHash, repo, logger)
+					if err != nil {
+						return err
+					}
```
juliev0, line 181: looks good but can you move the call to `ApplyPatchToResources()` above `updateCommitStatus()`? In the case the Controller goes down, we only want to mark this as committed in our Status once it's truly been committed

shubhamdixit863: ok done

shubhamdixit863: > Also, you need to set `lastCommitHash` to `recentHash`

yeah sorry forgot that

internal/git/gitsync_processor.go
```diff
@@ -106,7 +106,7 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, rest
 		logger.Errorw("error resolving the revision", "revision", repo.TargetRevision, "err", err, "repo", repo.RepoUrl)
 		return err
 	}
-
+	lastCommitHash = hash.String()
```
juliev0, line 109: I think we need to set `lastCommitHash` after this line below:
```
val, ok := gitSync.Status.CommitStatus[repo.Name]
```
We should add for the `ok` case:
```
lastCommitHash = gitSync.Status.CommitStatus[repo.Name].Hash
```

juliev0: This is the case where the Controller just restarted, and we get the last hash from the CommitStatus entry

shubhamdixit863: And what if we donot have any entry ? or is it the very first time cloning an syncing @juliev0

juliev0: I'm also thinking it logically makes sense for readability to move this line to the `!ok` case after we apply the resources because that's where the `lastCommitHash` truly is the last one set.

internal/git/gitsync_processor.go
```diff
@@ -122,6 +121,7 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, rest
 		}
 	}
 	val, ok := gitSync.Status.CommitStatus[repo.Name]
+	lastCommitHash = gitSync.Status.CommitStatus[repo.Name].Hash
```
juliev0, line 124: this should only be in the case of `ok==true`

juliev0: and in the case of `ok==false`, we still need to set it there after we apply the resources (to the hash)

internal/git/gitsync_processor.go
juliev0: > And what if we donot have any entry ? or is it the very first time cloning an syncing @juliev0

that's the `!ok` case I mention above

internal/git/gitsync_processor.go
```diff
@@ -168,21 +171,21 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1.GitSync, rest
 			select {
 			case <-ticker.C:
 				logger.Debug("checking for new updates")
-				patchedResources, recentHash, err := CheckForRepoUpdates(ctx, r, repo, &gitSync.Status, namespace)
-				if err != nil {
-					return err
-				}
-
-				err = updateCommitStatus(ctx, k8Client, namespacedName, recentHash, repo, logger)
+				patchedResources, recentHash, err := CheckForRepoUpdates(ctx, r, repo, lastCommitHash, namespace)
```
juliev0, line 174: now we need to change this `lastCommitHash` back to `hash.String()`

xdevxy:
Verified in a local cluster, no longer saw the error messages again.

juliev0 (approved):

Issue #22, opened 2024-01-11
add Validation logic for GitSync 


juliev0:
Need to make sure the GitSync CRD that's applied is valid (there's a placeholder - an empty validate() method - for it in the Controller).

Basically, need to go through the expected fields of the Spec and make sure they are populated in a valid way.

shubhamdixit863:
Following Validations need to be checked 
1-repourl is valid or not by running git ls-remote using Go git library
2-Names can only contain lowercase alphanumeric characters, '-', and '.', but must start and end with an alphanumeric character. They cant contain names starting with kubernetes-, kube-, as these are reserved names.
3-targetRevision cannot be an empty string 
4-name for repositoryPaths cannot be empty string 
5-clusterName cannot be empty string
@juliev0 @xdevxy

xdevxy:
Closed by #74

PR #74, opened 2024-02-01
Validate gitsync crd

shubhamdixit863:
https://github.com/numaproj-labs/numaplane/issues/22


internal/controller/gitsync_controller.go
```diff
@@ -273,9 +275,38 @@ func (r *GitSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			}
+
+			if !validations.IsValidName(destinations[i].Namespace) {
+				return fmt.Errorf("namespace is not a valid string for cluster %s", destinations[i].Cluster)
+			}
+		}
+		return nil
+	}
+	return errors.New("repositoryPaths and destinations cannot be empty")
+
```
juliev0, line 309: make sure there's at least one RepositoryPath and at least one Destination

could maybe also verify that no 2 Destinations {cluster+namespace} are identical

internal/shared/validations/validations.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/go-git/go-git/v5/config"
+)
+
+// CheckGitURL checks whether the given git url is a valid git url or not by fetching remote references
+// TODO: for private repository
+func CheckGitURL(gitURL string) bool {
+	rem := gogit.NewRemote(nil, &config.RemoteConfig{
+		Name: "origin",
+		URLs: []string{gitURL},
+	})
```
juliev0, line 16: hmm, I wonder if we want to do this....from the validation I've seen before, it's always fairly simple and there's no reaching out across the network. We also wouldn't want the case where some intermittent network error causes us to fail validation, which I think could happen here

shubhamdixit863: Yeah i see ,so we should not look for the valid github repo url and wait till user gets the error while cloning ? @juliev0

internal/shared/validations/validations.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	return true
+}
+
+func IsValidName(name string) bool {
+	// names can only contain lowercase alphanumeric characters, '-', and '.', but must start and end with an alphanumeric
+	validNameRegex := regexp.MustCompile(`^[a-z0-9][a-z0-9\-.]*[a-z0-9]$`)
+
+	//cant contain names starting with kubernetes-, kube-, as these are reserved names
+	reservedNamesRegex := regexp.MustCompile(`^(kubernetes-|kube-)`)
```
juliev0, line 31: good finding on this, but I think it only applies to namespaces and not repos, right? so maybe don't call the same method for both?

internal/controller/gitsync_controller.go
```diff
@@ -273,9 +275,38 @@ func (r *GitSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	return nil
+	// validate gitSync repositoryPaths
+	repositoryPaths := gitSync.Spec.RepositoryPaths
+	destinations := gitSync.Spec.Destinations
+	if len(repositoryPaths) > 0 && len(destinations) > 0 {
+		for i := 0; i < len(repositoryPaths); i++ {
+			if ok := validations.CheckGitURL(repositoryPaths[i].RepoUrl); !ok {
+				return fmt.Errorf("invalid remote repository url %s", repositoryPaths[i].RepoUrl)
+			}
+			// TODO : make this error message for informative
```
juliev0, line 287:
```suggestion
			// TODO : make this error message more informative
```

internal/shared/validations/validations.go
shubhamdixit863: ok ,i will remove the network part ,but still there should be some check to validate whether the git url is proper ,something like a regex check . As git url can be with https or git:// ,I was checking out this stackoverflow question -https://stackoverflow.com/questions/31801271/what-are-the-supported-git-url-formats ,should we implement something like this @juliev0

juliev0: yes, that sounds good - thanks

internal/shared/validations/validations.go
```diff
@@ -0,0 +1,68 @@
...
+	u, err := url.Parse(gitURL)
+	if err == nil && !Transports.Valid(u.Scheme) {
+		return false
+	}
+	return true
+}
+
+func IsValidName(name string) bool {
+	// names can only contain lowercase alphanumeric characters, '-', and '.', but must start and end with an alphanumeric
+	validNameRegex := regexp.MustCompile(`^[a-z0-9][a-z0-9\-.]*[a-z0-9]$`)
```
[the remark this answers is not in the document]
juliev0: By the way, how did you determine this definition of names: "can only contain lowercase alphanumeric characters, '-', and '.', but must start and end with an alphanumeric"? Actually, maybe we should just allow any string?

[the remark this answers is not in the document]
juliev0: Oh, I see you're using this for both K8S namespace and name of RepositoryPath. So, maybe this is correct for K8S namespace? But we don't need this for RepositoryPath name. For that, probably just making sure it's not an empty string is sufficient.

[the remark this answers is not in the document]
juliev0: For namespaces I see [this](https://release.com/blog/10-kubernetes-namespace-best-practices-to-start-following#:~:text=Namespaces%20in%20Kubernetes%20follow%20the,characters%20can%20only%20be%20lowercase.): "You can create a name with a maximum length of 253 characters using only alphanumeric characters and hyphens. Names cannot start with a hyphen and the alpha characters can only be lowercase."

internal/controller/gitsync_controller.go
```diff
@@ -273,8 +276,47 @@ func (r *GitSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if len(repositoryPaths) == 0 || len(destinations) == 0 {
+		return errors.New("both repositoryPaths and destinations must have at least one item")
+	}
+
+	// Validate each repositoryPath
+	for i := range repositoryPaths {
+		if ok := validations.CheckGitURL(repositoryPaths[i].RepoUrl, ctx); !ok {
+			return fmt.Errorf("invalid remote repository url %s", repositoryPaths[i].RepoUrl)
+		}
+		if !validations.IsValidName(repositoryPaths[i].Name) {
```
juliev0, line 295: per my other comment, just check that it's not empty instead

shubhamdixit863: ok . Will put a check for empty string only

shubhamdixit863: Addressed

internal/shared/validations/validations.go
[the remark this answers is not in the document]
shubhamdixit863: Added 253 chars length too

internal/shared/validations/validations.go
```diff
@@ -0,0 +1,71 @@
...
+
+func CheckGitURL(gitURL string, ctx context.Context) bool {
+	u, err := url.Parse(gitURL)
+	if err == nil && !Transports.Valid(u.Scheme) {
+		return false
+	}
+	return true
+}
+
+func IsValidName(name string) bool {
```
juliev0, line 58:
```suggestion
func IsValidKubernetesNamespace(name string) bool {
```

internal/controller/gitsync_controller.go
```diff
@@ -273,8 +276,47 @@ func (r *GitSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		key := fmt.Sprintf("%s-%s", dest.Cluster, dest.Namespace)
+		if _, exists := destinationSet[key]; exists {
+			return fmt.Errorf("duplicate destination found for cluster %s and namespace %s", dest.Cluster, dest.Namespace)
+		}
+		destinationSet[key] = struct{}{}
+
+		if len(dest.Cluster) == 0 {
+			return fmt.Errorf("cluster name cannot be empty")
+		}
+		if !validations.IsValidName(dest.Namespace) || validations.IsReservedName(dest.Namespace) {
```
juliev0, line 315:
```suggestion
		if !validations.IsValidKubernetesNamespace(dest.Namespace) || validations.IsReservedName(dest.Namespace) {
```

internal/shared/validations/validations.go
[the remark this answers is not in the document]
juliev0: shouldn't "[a-z0-9\-.]" remove the period at the end?

config/manager/manager.yaml
```diff
@@ -74,6 +74,7 @@ spec:
           - name: CLUSTER_NAME
             value: CLUSTER_NAME_VALUE # Replace it with the cluster name where numaplane will get deployed.
         image: numaplane-controller:latest
+        imagePullPolicy: IfNotPresent
```
xdevxy, line 77: Why do you adding this configuration? I think it is good to pull each time to avoid local image is out dated.

internal/shared/validations/validations.go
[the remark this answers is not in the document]
whynowy: Please note there are already many existing validation utils provided in k8s.io/apimachinery/pkg/util/validation/field.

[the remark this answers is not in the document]
shubhamdixit863: @whynowy I looked into the methods ,the nearest I found is `validation.IsValidLabelValue()` which is as per this -https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
But for namespaces we don't have and method for validation as it should follow RFC 112 standard -https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
Mentioned here -https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#namespaces-and-dns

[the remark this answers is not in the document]
shubhamdixit863: > shouldn't "[a-z0-9-.]" remove the period at the end?
yes we dnt need periods

internal/shared/validations/validations.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	return true
+}
+
+func IsValidKubernetesNamespace(name string) bool {
+	if len(name) > 63 {
+		return false
+	}
+	// names can only contain lowercase alphanumeric characters, '-', but must start and end with an alphanumeric
+	validNameRegex := regexp.MustCompile(`^[a-z0-9][a-z0-9\-]*[a-z0-9]$`)
```
juliev0, line 63: I'm kind of thinking this should call `IsReservedName()`? I mean, it technically should fail to be considered a "valid namespace" if it's a reserved name, right?

shubhamdixit863: ok i will update it

shubhamdixit863: so do you want me to merge both methods?

internal/shared/validations/validations.go
```diff
@@ -0,0 +1,71 @@
...
+
+func CheckGitURL(gitURL string, ctx context.Context) bool {
+	u, err := url.Parse(gitURL)
+	if err == nil && !Transports.Valid(u.Scheme) {
+		return false
+	}
+	return true
+}
+
+func IsValidKubernetesNamespace(name string) bool {
```
xdevxy, line 58: Please use the built in validation for k8s package. https://github.com/kubernetes/kubernetes/blob/2fb99eb9e81ec9fc92d8e694227772928f2eecc5/staging/src/k8s.io/pod-security-admission/admission/api/validation/validation.go#L73C20-L73C41

internal/shared/validations/validations.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(name) > 63 {
+		return false
+	}
+	// names can only contain lowercase alphanumeric characters, '-', but must start and end with an alphanumeric
+	validNameRegex := regexp.MustCompile(`^[a-z0-9][a-z0-9\-]*[a-z0-9]$`)
+	return validNameRegex.MatchString(name)
+}
+
+func IsReservedName(name string) bool {
```
xdevxy, line 67: Can you also find is there any built in support to check Reserved name as well? Thanks!

internal/controller/gitsync_controller.go
```diff
@@ -273,8 +276,47 @@ func (r *GitSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-// TODO: add validation
 func (r *GitSyncReconciler) validate(gitSync *apiv1.GitSync) error {
+	repositoryPaths := gitSync.Spec.RepositoryPaths
+	destinations := gitSync.Spec.Destinations
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
xdevxy, line 282: Why do we need `context.WithTimeout`? We are not doing any network request or database request here.

internal/shared/validations/validations.go
shubhamdixit863: The above method internally uses this method -https://github.com/kubernetes/kubernetes/blob/2fb99eb9e81ec9fc92d8e694227772928f2eecc5/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L188 

And as per the k8 docs.  **As a result, all namespace names must be valid [RFC 1123 DNS labels]**(https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names).   So I have modified it

internal/controller/gitsync_controller.go
shubhamdixit863: it was used previoulsy ,but not anymore ,removed

juliev0 (approved):

xdevxy (approved):

Issue #79, opened 2024-02-04
Numaplane Controller ConfigMap


juliev0:
Now that we are starting to have more configurability, we can employ a `ConfigMap` for the Numaplane Controller to read.  

We can have a `Config` struct similar to [this GlobalConfig struct in numaflow](https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/config.go#L29), which is read in by a method like [this](https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/config.go#L127). This file could go in our `internal/controller` directory. 

We can replace this code which looks for the CLUSTER_NAME in an environment variable [here](https://github.com/numaproj-labs/numaplane/blob/main/internal/controller/gitsync_controller.go#L59) with code which instead creates one of those `Config` structs and loads it [here](https://github.com/numaproj-labs/numaplane/blob/main/internal/controller/gitsync_controller.go#L59). The `*Config` could in fact be a member of the `GitSyncReconciler`.

This ConfigMap will include:
- cluster name
- time interval for watching repos ([this](https://github.com/numaproj-labs/numaplane/issues/77) issue)
- a mapping of RepositoryPathPrefix (e.g. "https://github.intuit.com/") to the credentials (i.e. a reference to the Kubernetes Secrets/keys within those Secrets (similar to [this](https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/redis_buffer_service.go#L50)))

See the Numaflow Controller ConfigMap for what [theirs](https://github.com/numaproj/numaflow/blob/main/config/base/controller-manager/numaflow-controller-config.yaml) looks like as a reference. You can create one locally for your testing which has all of the fields you want to use. 

[...]

Issue #77, opened 2024-02-01
`timeInterval` for watch a repo should be configurable


xdevxy:
The `timeInterval` defined [here](https://github.com/numaproj-labs/numaplane/blob/main/internal/git/gitsync_processor.go#L30) should be configurable.

xdevxy:
Closed by #80

PR #80, opened 2024-02-08
GlobalConfigMap For Controller

shubhamdixit863:
https://github.com/numaproj-labs/numaplane/issues/79
https://github.com/numaproj-labs/numaplane/issues/77 ,time interval is now configurable with the GlobalConfig


juliev0:
@whynowy One thing I noticed with this viper code borrowed from [Numaflow](https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/config.go#L143) is that technically you can reload it and it could be accessed elsewhere at the same time with no concurrency measures taken, right? Is that a calculated risk based on the rareness of this sort of race condition?

internal/controller/config.go
```diff
@@ -0,0 +1,42 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+// GlobalConfig is the configuration for the controllers, it is
+// supposed to be populated from the configmap attached to the
+// controller manager.
+type GlobalConfig struct {
+	ClusterName     string                               `yaml:"clusterName" json:"clusterName"`
+	TimeInterval    uint                                 `yaml:"timeInterval" json:"timeInterval"`
+	RepoCredentials map[string]*corev1.SecretKeySelector `json:"repoCredentials"`
```
juliev0, line 17: this one specifies json and not yaml - is that on purpose?

juliev0, line 17: I'm kind of thinking we should have a new struct called `GitCredential` and map to that. The `GitCredential` struct can consist of:
```
type GitCredential struct {
    httpCredential *HTTPCredential 
    sshCredential *SSHCredential
}

type HTTPCredential struct {
    Username string
    Password string
}
type SSHCredential struct {
   Name string
   SSHKey string
}
```

Let me get back to you on also incorporating TLS settings into this...

internal/controller/config.go
```diff
@@ -0,0 +1,42 @@
...
+}
```
juliev0, line 18: I think we should change the `RepoCredentials` a bit to be able to incorporate HTTP, SSH, and TLS - let me get back to you on that

juliev0: I'm thinking something like this:
```
type GitCredential struct {
    HTTPCredential *HTTPCredential 
    SSHCredential *SSHCredential
    TLS *TLS
}

type HTTPCredential struct {
    Username string
    Password *corev1.SecretKeySelector
}
type SSHCredential struct {
   Name string
   SSHKey *corev1.SecretKeySelector
}

type TLS struct {
	// +optional
	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty" protobuf:"bytes,1,opt,name=insecureSkipVerify"`
	// CACertSecret refers to the secret that contains the CA cert
	// +optional
	CACertSecret *corev1.SecretKeySelector `json:"caCertSecret,omitempty" protobuf:"bytes,2,opt,name=caCertSecret"`
	// CertSecret refers to the secret that contains the cert
	// +optional
	CertSecret *corev1.SecretKeySelector `json:"clientCertSecret,omitempty" protobuf:"bytes,3,opt,name=certSecret"`
	// KeySecret refers to the secret that contains the key
	// +optional
	KeySecret *corev1.SecretKeySelector `json:"clientKeySecret,omitempty" protobuf:"bytes,4,opt,name=keySecret"`
}
```

juliev0: What I'm thinking we'll put in the key for the `map[string]*GitCredential` is something like "github.intuit.com". You can imagine that somebody may specify a RepositoryPath like "https://github.intuit.com/myrepo" or "ssh://github.intuit.com/myrepo". We could specify both the SSHCredential there as well as the HTTPCredential, assuming that credentials are required for each of them. And the TLS config would apply to both of them.

juliev0: (sorry if we may need to iterate on what I have above at some point - this is what I'm thinking at the moment

whynowy:
> @whynowy One thing I noticed with this viper code borrowed from [Numaflow](https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/config.go#L143) is that technically you can reload it and it could be accessed elsewhere at the same time with no concurrency measures taken, right? Is that a calculated risk based on the rareness of this sort of race condition? (where rareness depends on how often it's read in elsewhere I guess)

You are right, the code in Numaflow is buggy, check [this one](https://github.com/numaproj-labs/log-ingestor/blob/main/pkg/ingestor/ingestor.go#L168) instead.

internal/controller/config.go
shubhamdixit863: `*corev1.SecretKeySelector`  has an embedded struct  -```
type SecretKeySelector struct {
	// The name of the secret in the pod's namespace to select from.
	LocalObjectReference `json:",inline" protobuf:"bytes,1,opt,name=localObjectReference"`
	// The key of the secret to select from.  Must be a valid secret key.
	Key string `json:"key" protobuf:"bytes,2,opt,name=key"`
	// Specify whether the Secret or its key must be defined
	// +optional
	Optional *bool `json:"optional,omitempty" protobuf:"varint,3,opt,name=optional"`
}
```
because of which name is not able to get parsed by viper `Loadconfig` which is inside the `LocalObjectReference `struct

```
type LocalObjectReference struct {
	// Name of the referent.
	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
	// TODO: Add other useful fields. apiVersion, kind, uid?
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
}
```
We can use `corev1.SecretKeySelector` if we dnt require name key otherwise its limitation with viper -see this https://github.com/spf13/viper/issues/797
I can replicate corev1.SecretKeySelector and can add the `mapstructure:",squash"` that might solve the problem

See working issue -https://replit.com/join/qkzwkrlabn-shubhamdixit863

Comment (juliev0) [reply internal/controller/config.go]: Great research. 
> I can replicate corev1.SecretKeySelector and can add the `mapstructure:",squash"` that might solve the problem

That sounds good I think!

Comment (juliev0) [reply internal/controller/config.go]: Let's make sure we're keeping viper - have you looked at this [concurrency](https://github.com/numaproj-labs/numaplane/pull/80#issuecomment-1934681589) issue at all? Do you know what I mean?

Comment (shubhamdixit863) [reply internal/controller/config.go]: yeah i understood the point as the callback for reload can change the shared Globalconfig that might cause the race condition .I will initialize the mutex and and add a lock probably here . I think RW lock would be better in this scenario as we will have less frequent writes as compared to reads what do you think @juliev0

File: internal/controller/config/config.go [hunk 3]
```diff
@@ -0,0 +1,70 @@
...
+	}
+	v.WatchConfig()
+	v.OnConfigChange(func(e fsnotify.Event) {
+		err = v.Unmarshal(r)
+		if err != nil {
+			onErrorReloading(err)
+		}
+	})
+	return r, nil
+}
```
juliev0, line 70: I'm thinking that the way to deal with [this](https://github.com/numaproj-labs/numaplane/pull/80#issuecomment-1934681589) concurrency issue may be to create a struct called `ConfigManager` which holds the official copy of the `GlobalConfig`. At any time somebody can call `func (cm *ConfigManager) getConfig() GlobalConfig`, which you'll notice returns a copy of the `GlobalConfig` which is guaranteed to be atomically correct. Any user should call this once to get the `Config` and use it to get the various fields from it.

`ConfigManager` can have a `GlobalConfig` as a member, which is what it would return when `GetConfig()` is called. 
This `LoadConfig()` function could be a member of the `ConfigManager` struct. It should no longer return a `*GlobalConfig`. Instead it will simply set its own `GlobalConfig` member. 
`v.OnConfigChange()` will set the member as well. The `v.OnConfigChange()` code block and the `GetConfig()` method will need to both use the same mutex

shubhamdixit863: ok got it .I will make the changes

juliev0: So, to summarize, something like this:
```
type ConfigManager struct {
   config GlobalConfig
   lock sync.RWMutex
}
func (cm *ConfigManager) GetConfig() GlobalConfig {
   lock.RLock()
   defer lock.RUnlock()
   return config
}
func (cm *ConfigManager) LoadConfig() error {
    // add logic to set 'config' here
    ....
    v.OnConfigChange(func(e fsnotify.Event) {
                lock.WLock()
                defer lock.WUnlock()
		// add logic to set 'config' here
	})
}

```

whynowy: This is right. https://github.com/numaproj-labs/log-ingestor/blob/main/pkg/ingestor/ingestor.go#L168

juliev0: Oh, I guess you already thought about this issue. :) In our case, I also don't want the user to hold a reference to the config either: otherwise they might for example get the Username field and then the Password field, and in between the reference got changed so it's a mismatch.

whynowy:
https://github.com/numaproj/numaflow/pull/1499

internal/controller/config/config.go
```diff
@@ -0,0 +1,90 @@
...
+type SecretKeySelector struct {
+	corev1.LocalObjectReference `mapstructure:",squash"` // for viper to correctly parse the config
+	Key                         string                   `json:"key" `
+	Optional                    *bool                    `json:"optional,omitempty" `
+}
+
+func (cm *ConfigManager) GetConfig() *GlobalConfig {
+	cm.lock.RLock()
+	defer cm.lock.RUnlock()
+	return cm.config
```
juliev0, line 64: What I was thinking was that this should return a `GlobalConfig` as opposed to the pointer. If it returns the pointer, then it's returning a reference to what's stored within the `ConfigManager`, which could change on it, when it's in the middle of reading it. 

The idea is that any user of this would get a copy of the `GlobalConfig`, guaranteed to be atomically correct, periodically, if it needs to. Then it's guaranteed to be regularly refreshing to the latest, but at the same time getting a copy that's atomically correct.

internal/controller/config/config.go
```diff
@@ -0,0 +1,90 @@
...
+		return fmt.Errorf("failed to load configuration file. %w", err)
+	}
+	err = v.Unmarshal(cm.config)
+	if err != nil {
+		return fmt.Errorf("failed unmarshal configuration file. %w", err)
+	}
+	v.WatchConfig()
+	v.OnConfigChange(func(e fsnotify.Event) {
+		cm.lock.RLock()
+		defer cm.lock.RUnlock()
```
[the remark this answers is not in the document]
juliev0: (sorry, my bad that I originally wrote `WLock` and `WUnlock` in my other comment. I should've looked up the actual syntax.) In any case, this is the writer. (You can look up RWMutex if it's a new concept. It's optimized to allow multiple readers to read simultaneously, but a reader and writer can't be concurrent, and neither can 2 writers.)

cmd/main.go
```diff
@@ -96,9 +98,18 @@ func main() {
...
+	err = configManager.LoadConfig(func(err error) {
+		logger.Errorw("Failed to reload global configuration file", err)
+	}, configPath)
+	if err != nil {
+		logger.Fatalw("Failed to load config file", err)
+	}
 	reconciler, err := controller.NewGitSyncReconciler(
 		kubeClient,
 		mgr.GetScheme(),
+		configManager.GetConfig(),
```
juliev0, line 112: let's pass the `configManger` in instead. That should be what gets passed in everywhere. Then, the user who needs updated configuration can always call `configManager.GetConfig()` whenever it needs an atomically correct version of the configuration.

internal/git/gitsync_processor.go
```diff
@@ -434,7 +458,7 @@ func getLatestCommitHash(repo *git.Repository, refName string) (*plumbing.Hash,
 	return commitHash, err
 }
 
-func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeClient kubernetes.Client, clusterName string) (*GitSyncProcessor, error) {
+func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeClient kubernetes.Client, clusterName string, repoCred map[string]*controllerconfig.GitCredential) (*GitSyncProcessor, error) {
```
juliev0, line 461: Sorry, I know my original thought was to just pass in the credentials map here, but now I'm thinking that either we pass in the `configManager`, or maybe even better we could just have a singleton ConfigManager that anyone can access...

internal/controller/config/config_test.go
```diff
@@ -0,0 +1,126 @@
...
+	assert.Equal(t, "ca-cert-secret", config.RepoCredentials["key3"].TLS.CACertSecret.Name, "CACertSecret Name for TLS does not match")
+	assert.Equal(t, "ca.crt", config.RepoCredentials["key3"].TLS.CACertSecret.Key, "CACertSecret Key for TLS does not match")
+	assert.Equal(t, "cert-secret", config.RepoCredentials["key3"].TLS.CertSecret.Name, "CertSecret Name for TLS does not match")
+	assert.Equal(t, "tls.crt", config.RepoCredentials["key3"].TLS.CertSecret.Key, "CertSecret Key for TLS does not match")
+	assert.Equal(t, "key-secret", config.RepoCredentials["key3"].TLS.KeySecret.Name, "KeySecret Name for TLS does not match")
+	assert.Equal(t, "tls.key", config.RepoCredentials["key3"].TLS.KeySecret.Key, "KeySecret Key for TLS does not match")
+}
+
+// to verify this test run with go test -race ./... it  won't give a race condition as we have used mutex.RwLock
+// in onConfigChange in LoadConfig
```
juliev0, line 52: neat idea to test this for the race condition. I wonder if we should in that case update our [Makefile](https://github.com/numaproj-labs/numaplane/blob/main/Makefile#L75) so we have this flag

juliev0: [Here's](https://github.com/open-policy-agent/opa/issues/2388) somebody else who thought about the same thing but decided against it due to it slowing the build down. I wonder if maybe we should add it, and if we have an issue with build time we could remove it. 

(Alternative would be to either add a different Makefile target `make test-with-race` or something, or alternatively have a `TEST_FLAGS` variable that could be set when running, i.e.:
```
        test: manifests generate fmt vet envtest ## Run tests.
	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -v ${TEST_FLAGS} ./... -coverprofile cover.out
```

internal/controller/config/config.go
```diff
@@ -0,0 +1,97 @@
...
+	return instance
+}
+
+// GlobalConfig is the configuration for the controllers, it is
+// supposed to be populated from the configmap attached to the
+// controller manager.
+type GlobalConfig struct {
+	ClusterName     string                    `json:"clusterName"`
+	TimeIntervalSec uint                      `json:"timeIntervalSec"`
+	RepoCredentials map[string]*GitCredential `json:"repoCredentials"`
```
juliev0, line 37:
```suggestion
        // RepoCredentials maps each Git Repository Path prefix to the corresponding credentials that are needed for it
	RepoCredentials map[string]*GitCredential `json:"repoCredentials"`
```

config/manager/basicsecret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: secret-basic-auth
+  namespace: numaplane-system
+type: kubernetes.io/basic-auth
+stringData:
+  username: "admin" # required field for kubernetes.io/basic-auth
+  password: "t0p-Secret" # required field for kubernetes.io/basic-auth
```
juliev0, line 9: this Secret is maybe okay to go in the `samples` directory. Let's not put it in here, since this is what will be deployed by default.

config/manager/config.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: numaplane-controller-config
+  namespace: numaplane-system
+data:
+  config.yaml: |
+    clusterName: "example-cluster"
+    timeIntervalSec: 60
+    repoCredentials:
```
juliev0, line 10: Let's take the `repoCredentials` out of here. This represents our default manifest which we will deploy. 

To be honest, I think we ultimately will need to even remove `clusterName` from here and expect that that will be added by the kustomization which is done outside of this repo, but for now let's make it so that [this](https://github.com/numaproj-labs/numaplane/blob/main/Makefile#L149) line works. See how it's configured as an environment variable in the `Deployment` object [here](https://github.com/numaproj-labs/numaplane/blob/main/config/manager/manager.yaml#L75)? Let's remove it from there and instead set `clusterName: CLUSTER_NAME_VALUE` here

config/samples/config.yaml
```diff
@@ -1,6 +1,6 @@
 clusterName: "example-cluster"
 timeIntervalSec: 60
-repoCredentials:
+epoCredentials:
```
juliev0, line 3: this got messed up

config/manager/kustomization.yaml
```diff
@@ -1,5 +1,7 @@
 resources:
 - manager.yaml
+- config.yaml
+- ../samples/basicsecret.yaml
```
juliev0, line 4: no, don't include the Secret here at all. This kustomization.yaml represents what Numaplane will install by default.

config/manager/config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: numaplane-controller-config
+  namespace: numaplane-system
+data:
+  config.yaml: |
+    clusterName: "CLUSTER_NAME_VALUE" # Replace with the cluster name where numaplane will get deployed.
```
juliev0, line 8: Thanks. Did you test `make deploy` to confirm that it gets correct set?

config/samples/config.yaml
```diff
@@ -0,0 +1,26 @@
...
+      insecureSkipVerify: true
+      CACertSecret:
+        name: "ca-cert-secret"
+        key: "ca.crt"
+      CertSecret:
+        name: "cert-secret"
+        key: "tls.crt"
+      KeySecret:
+        name: "key-secret"
+        key: "tls.key"
```
juliev0, line 26: So, I think you said something yesterday about actually testing with public/private key or something? Were you referring to TLS then?

config/samples/config.yaml
```diff
@@ -0,0 +1,26 @@
...
+        name: "http-creds"
+        key: "password"
+  "key2":
+    SSHCredential:
+      SSHKey:
+        name: "ssh-creds"
+        key: "sshKey"
+  "key3":
+    TLS:
+      insecureSkipVerify: true
```
juliev0, line 17: I'm kind of thinking this should be false in your example? I think it means that we don't try to verify the server side's identity. If we have all of those files listed below, then we probably are verifying their identity.

internal/controller/config/config.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type SSHCredential struct {
+	SSHKey SecretKeySelector `json:"SSHKey" yaml:"SSHKey" `
+}
+
+type TLS struct {
+	InsecureSkipVerify bool              `json:"insecureSkipVerify"`
+	CACertSecret       SecretKeySelector `json:"CACertSecret"`
+	CertSecret         SecretKeySelector `json:"CertSecret"`
```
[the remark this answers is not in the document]
juliev0: (not an acronym like the other ones starting with upper case)

internal/controller/config/config.go
```diff
@@ -0,0 +1,98 @@
...
+
+type SecretKeySelector struct {
+	corev1.LocalObjectReference `mapstructure:",squash"` // for viper to correctly parse the config
+	Key                         string                   `json:"key" `
+	Optional                    *bool                    `json:"optional,omitempty" `
+}
+
+func (cm *ConfigManager) GetConfig() *GlobalConfig {
+	cm.lock.Lock()
+	defer cm.lock.Unlock()
```
juliev0, line 71:
```suggestion
	cm.lock.RLock()
	defer cm.lock.RUnlock()
```

juliev0: (this is a reader.....using this function allows multiple readers to read at the same time, which is deemed safe)

internal/git/gitsync_processor.go
```diff
@@ -74,9 +79,27 @@ type MetaData struct {
...
+	secret := &corev1.Secret{}
+	key := k8sClient.ObjectKey{
+		Namespace: namespace,
+		Name:      secretName,
+	}
+	if err := kubeClient.Get(ctx, key, secret); err != nil {
+		return nil, err
+	}
+	return secret, nil
+}
```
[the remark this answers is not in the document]
juliev0: @chandankumar4 I thought maybe we could move this general-purpose function into `internal/kubernetes` directory somewhere. Any recommendation?

config/manager/config.yaml
shubhamdixit863: Removing this will cause the loadConfig fail as it needs all the keys .

config/samples/config.yaml
shubhamdixit863: Picking this up in follow up pr

config/manager/config.yaml
shubhamdixit863: yes if you set the proper value in clusterName something like numaplane-cluster ,it works

internal/git/gitsync_processor.go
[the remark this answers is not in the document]
chandankumar4: Yes, we can move it inside `internal/kubernetes/kubernetes.go`, currently there is no such file called `kubernetes.go` but we can rename `deploy.go` to it as it's serving more purpose than just deploy.

config/manager/manager.yaml
```diff
@@ -70,10 +70,8 @@ spec:
         - /manager
         args:
         - --leader-elect
-        env:
-          - name: CLUSTER_NAME
-            value: CLUSTER_NAME_VALUE # Replace it with the cluster name where numaplane will get deployed.
         image: numaplane-controller:latest
+        imagePullPolicy: IfNotPresent
```
juliev0, line 74: did this get changed by accident after local testing?

config/samples/config.yaml
juliev0: any reason you don't want to change it here?

internal/git/gitsync_processor.go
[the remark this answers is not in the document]
juliev0: can you do this @shubhamdixit863 ?

Dockerfile
```diff
@@ -28,7 +28,7 @@ RUN --mount=type=cache,target="/root/.cache/go-build" CGO_ENABLED=0 GOOS=${TARGE
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM alpine:latest
```
juliev0, line 31: did this get changed by accident?

config/samples/config.yaml
juliev0: I know it's a little crazy making for you to be working on these two PRs at the same time. At the same time, I don't want PR comments to get lost, so in general it would be good if we can address within the PR

internal/controller/config/config.go
juliev0: Why was this resolved @shubhamdixit863 ?

Dockerfile
shubhamdixit863: yeah i was testing for the presence of  config yaml  in the internal folder by executing shell command . Somehow it was missed :(.

config/manager/manager.yaml
shubhamdixit863: yes, I definitely need a new file for local testing .

internal/controller/config/config.go
shubhamdixit863: But we would be needing a Deep copy method then ,as internal values are also pointers ? @juliev0 ,may be like a nested deep copy  ?

juliev0: > But we would be needing a Deep copy method then ,as internal values are also pointers ? @juliev0 ,may be like a nested deep copy ?

great call out there, yeah

shubhamdixit863: Yeah ,ok I have  to restructure controller tests too ,then.

Makefile
```diff
@@ -75,6 +75,12 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -v ./... -coverprofile cover.out
 
+
+.PHONY: test-with-race
+test-with-race: manifests generate fmt vet envtest ## Run tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -race ./... -coverprofile cover.out
+
+
```
juliev0, line 83:
```suggestion
test: manifests generate fmt vet envtest ## Run tests.
	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -race -v ./... -coverprofile cover.out

```

juliev0: Sorry, now I'm thinking we just put "-race" flag in the main test target which will be run by the CI. I noticed numaflow is including it in their test

internal/git/gitsync_processor.go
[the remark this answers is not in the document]
shubhamdixit863: addressed in this pr -https://github.com/numaproj-labs/numaplane/pull/94 ,as I require it there

xdevxy:
> @xdevxy I told Shubham that we could merge this without the DCO being fixed, since it sounds like it will be a pain to rebase.

Sounds good.

Makefile
```diff
@@ -73,7 +73,9 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -v ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -race -v ./... -coverprofile cover.out
```
xdevxy, line 76: 👍

config/samples/basicsecret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: secret-basic-auth
+  namespace: numaplane-system
+type: kubernetes.io/basic-auth
+stringData:
+  username: "admin" # required field for kubernetes.io/basic-auth
+  password: "t0p-Secret" # required field for kubernetes.io/basic-auth
```
xdevxy, line 9: nit: newline at the end.

xdevxy (changes requested):
Please do not commit `internal/.DS_Store` file.

shubhamdixit863:
> please add .DS_STORE to .gitignore

already added in main branch

juliev0:
@xdevxy since you said you'd approve once the .DS_STORE file was removed, I merged it. Hope that works.

PR #86, opened 2024-02-13
feat: Add support to deploy kustomized enabled repo

chandankumar4:
This commit will add feature to deploy manifest from kustomize enabled repo.


xdevxy (commented):
Overall LGTM, just want to make sure we have test coverage for kustomize manifests. Thanks!

internal/git/gitsync_processor.go
```diff
@@ -114,27 +129,46 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1alpha1.GitSync
 	val, ok := gitSync.Status.CommitStatus[repo.Name]
 	// Only create the resources for the first time if not created yet.
 	// Otherwise, monitoring with intervals.
+
+	repoAbsolutePath := localRepoPath + "/" + repo.Path
+	k := kustomize.NewKustomizeApp(repoAbsolutePath, repo.RepoUrl, "")
```
juliev0, line 134: nit: can this line go below where the kustomization logic is, or does it need to be over here? (seems more readable over there)

internal/git/gitsync_processor.go
```diff
@@ -156,16 +190,29 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1alpha1.GitSync
 			return err
 		}
 	} else {
-		ticker := time.NewTicker(timeInterval * time.Minute)
+		ticker := time.NewTicker(10 * time.Second)
```
juliev0, line 193: did this accidentally get changed for testing maybe?

internal/git/gitsync_processor.go
```diff
@@ -226,6 +273,64 @@ func ApplyPatchToResources(patchedResources PatchedResource, kubeClient kubernet
...
+		return patchedResources, recentHash, err
+	}
+
+	remoteRef, err := getLatestCommitHash(r, repo.TargetRevision)
+	if err != nil {
+		logger.Errorw("failed to get latest commits in the github repo", "err", err)
+		return patchedResources, recentHash, err
+	}
+
+	// If remote commit hash and local one is not equal, it means there is new changes made in remote repository.
```
juliev0, line 300:
```suggestion
	// If remote commit hash and local one aren't equal, it means there are new changes made in remote repository.
```

internal/git/gitsync_processor.go
```diff
@@ -444,17 +566,23 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeCli
 		channels:    channels,
 		clusterName: clusterName,
 	}
+
 	for _, repo := range gitSync.Spec.RepositoryPaths {
 		gitCh := make(chan Message, messageChanLength)
 		channels[repo.Name] = gitCh
+		// get repository name from repo url, which will be used to create dynamic local path for clone, i'e(gitSync-name + repo-name)
```
juliev0, line 573:
```suggestion
		// get repository name from repo url, which will be used to create dynamic local path for clone, i.e. (gitSync-name + repo-name)
```

internal/git/gitsync_processor.go
```diff
@@ -467,6 +595,34 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeCli
...
+	if localRepoPath != "" {
+		return fmt.Sprintf("%s/%s-%s", localRepoPath, gitSyncName, repoName)
+	} else {
+		return fmt.Sprintf("/tmp/%s-%s", gitSyncName, repoName)
+	}
+}
+
+// getRepositoryName parse the url and return the repo name from it.
+func getRepositoryName(url string) (string, error) {
+	// Remove protocol and hostname
```
juliev0, line 610: sorry, I think I didn't clarify what I meant when I said that we should use the GitSync name to RepositoryPath name. By "Repository Path name", I was simply referring to this field: https://github.com/numaproj-labs/numaplane/blob/main/api/v1alpha1/gitsync_types.go#L73

internal/git/gitsync_processor.go
chandankumar4: I have put it here, so that I don't have to define it again for infinite for loop in line [here](https://github.com/numaproj-labs/numaplane/pull/86/files#diff-62540bd8eefb6e01f31efd127880279741d18ecc6845bc8f1f05a438db36a0b1R205)

internal/git/gitsync_processor.go
chandankumar4: ohh, my bad!! Let me revert it

internal/git/gitsync_processor.go
chandankumar4: okay, Sorry I meant it wrong!! I'll make this change accordingly.

chandankumar4: But we need to make the name field as mandatory right? I don't think currently it is.

juliev0: well, there's [validation](https://github.com/numaproj-labs/numaplane/blob/main/internal/controller/gitsync_controller.go#L292) for it

chandankumar4: ohh, Yaa remember now that shubham added it, I was checking `gitsync_processor_test.go` tests and seen that many tests are missing `name` so thought validation is not there, but validation is in reconsile method

internal/git/gitsync_processor.go
```diff
@@ -156,16 +190,29 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1alpha1.GitSync
...
 			case <-ticker.C:
 				logger.Debug("checking for new updates")
-				patchedResources, recentHash, err := CheckForRepoUpdates(ctx, r, repo, lastCommitHash, namespace)
-				if err != nil {
-					return err
+				var (
+					recentHash       string
+					patchedResources PatchedResource
+				)
+				// TODO: Add switch case based on type(Kustomize/Helm/Yaml) of repo, Once type field is available in GitSync CR.
```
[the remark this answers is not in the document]
chandankumar4: Created, https://github.com/numaproj-labs/numaplane/issues/90
need to groom this issue and add more detail in it.

internal/git/gitsync_processor.go
juliev0: I guess one thing I should mention is that @xdevxy has a PR open right now which will remove the multiple RepositoryPath concept from a GitSync, so really we won’t end up needing to add the RepositoryPath name at all and the GitSync name will be sufficient for uniqueness

internal/git/gitsync_processor_test.go
```diff
@@ -713,7 +714,7 @@ func Test_watchRepo(t *testing.T) {
 			client.EXPECT().ApplyResource(gomock.Any(), testNamespace).AnyTimes()
 			client.EXPECT().StatusUpdate(ctx, gomock.Any()).AnyTimes()
 
-			watchErr := watchRepo(ctx, r, tc.gitSync, client, repo, testNamespace)
+			watchErr := watchRepo(ctx, r, tc.gitSync, client, repo, testNamespace, localRepoPath)
```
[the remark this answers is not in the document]
chandankumar4: Added

internal/git/gitsync_processor.go
chandankumar4: Sure, I have update the path generation method

config/manager/manager.yaml
```diff
@@ -97,7 +97,7 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 128Mi
+            memory: 500Mi
```
juliev0, line 100: Did you find that with no PVC, the added files do in fact get included in this memory calculation? Is that why you adjusted it?

chandankumar4: Sort off, while testing I found out that pod was crashing with OOM and seen this issue opened already https://github.com/src-d/go-git/issues/447, that's the reason I have increased it

internal/controller/gitsync_controller_test.go
```diff
@@ -33,13 +33,13 @@ var (
...
 					RepoUrl:        "https://github.com/numaproj-labs/numaplane-control-manifests.git",
-					Path:           "./numaflowController/",
+					Path:           "staging-usw2-k8s",
 					TargetRevision: "main",
 				},
 				{
 					Name:           "my-pipelines",
 					RepoUrl:        "https://github.com/numaproj-labs/numaplane-control-manifests.git",
-					Path:           "./pipelines/",
+					Path:           "staging-usw2-k8s",
```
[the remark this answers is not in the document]
chandankumar4: There is no such path "./pipelines/" in given repo url, so have given the correct path

internal/git/gitsync_processor.go
```diff
@@ -114,27 +129,47 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1alpha1.GitSync
 	val, ok := gitSync.Status.CommitStatus[repo.Name]
 	// Only create the resources for the first time if not created yet.
 	// Otherwise, monitoring with intervals.
+
+	// kustomizePath will the path of clone repository + the path where kustomize file is present.
```
juliev0, line 133:
```suggestion
	// kustomizePath will be the path of clone repository + the path where kustomize file is present.
```

internal/git/gitsync_processor.go
```diff
@@ -114,27 +129,47 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1alpha1.GitSync
...
+	kustomizePath := localRepoPath + "/" + repo.Path
+	k := kustomize.NewKustomizeApp(kustomizePath, repo.RepoUrl, os.Getenv("KUSTOMIZE_BINARY_PATH"))
```
juliev0, line 135: Is this environment variable optional?

config/manager/manager.yaml
juliev0: I wonder if we want to potentially remove the limit altogether here, and assume anyone who kustomizes outside of this file will set it to whatever they need (if they set it at all)

chandankumar4: Yes, while cloning it's consumes more memory

internal/controller/gitsync_controller_test.go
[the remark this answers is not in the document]
juliev0: oh, maybe the layout of that repo got changed? Any idea why the test didn't fail previously? Was it not looking at that field?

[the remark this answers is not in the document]
chandankumar4: yes, here it's not checking for path, it's just starting and reconcile and verify it running.

config/manager/manager.yaml
juliev0: I saw we remove this field. What do you think?

internal/git/gitsync_processor.go
```diff
@@ -226,6 +274,64 @@ func ApplyPatchToResources(patchedResources PatchedResource, kubeClient kubernet
...
+	}
+
+	remoteRef, err := getLatestCommitHash(r, repo.TargetRevision)
+	if err != nil {
+		logger.Errorw("failed to get latest commits in the github repo", "err", err)
+		return patchedResources, recentHash, err
+	}
+
+	// If remote commit hash and local one aren't equal, it means there are new changes made in remote repository.
+	if remoteRef.String() != lastCommitHash {
```
juliev0, line 302: just to verify, if the kustomization.yaml references an upstream yaml from another repo, we won't detect that, right? I think what @xdevxy had in mind was that we recalculate the manifests every time, to prevent not seeing when that changed. But I'm okay worrying about this in a subsequent PR.

config/manager/manager.yaml
chandankumar4: sure, we can remove it

internal/git/gitsync_processor.go
chandankumar4: Right, currently we don't detect kustomization reference to another repo, I can create an issue if not already open.

juliev0: yes please, and probably assign it to yourself? 
And maybe we should compare the two options of either: a. watching the upstream repositories for changes; b. not watching but just performing the kustomization change every cycle

chandankumar4: Created it #100 i'll add more detail to it

juliev0 (commented):
Did you test this for the case that the kustomization.yaml references a file in the same repository but one or more levels up (i.e. `../something`)?

chandankumar4:
> Did you test this for the case that the kustomization.yaml references a file in the same repository but one or more levels up (i.e. `../something`)?

Yes, i have tested it with numaflow repo, which has the same behaviour, as gitsync example-

Url: https://github.com/numaproj/numaflow.git
Path: config/namespace-install

Here path contains the kustomization file which is referring resource as [../base](https://github.com/numaproj/numaflow/blob/main/config/namespace-install/kustomization.yaml#L5)

juliev0:
> > Did you test this for the case that the kustomization.yaml references a file in the same repository but one or more levels up (i.e. `../something`)?
> 
> Yes, i have tested it with numaflow repo, which has the same behaviour, as gitsync example-
> 
> Url: https://github.com/numaproj/numaflow.git Path: config/namespace-install
> 
> Here path contains the kustomization file which is referring resource as [../base](https://github.com/numaproj/numaflow/blob/main/config/namespace-install/kustomization.yaml#L5)

And did you try modifying something from `../base`?

chandankumar4:
> > > Did you test this for the case that the kustomization.yaml references a file in the same repository but one or more levels up (i.e. `../something`)?
> 
> > 
> 
> > Yes, i have tested it with numaflow repo, which has the same behaviour, as gitsync example-
> 
> > 
> 
> > Url: https://github.com/numaproj/numaflow.git Path: config/namespace-install
> 
> > 
> 
> > Here path contains the kustomization file which is referring resource as [../base](https://github.com/numaproj/numaflow/blob/main/config/namespace-install/kustomization.yaml#L5)
> 
> 
> 
> And did you try modifying something from `../base`? 

No haven't this scenario, but I believe it should work because modifying anything from ../base will change commit history so the controller will pick it up and apply it.

juliev0:
> > > > Did you test this for the case that the kustomization.yaml references a file in the same repository but one or more levels up (i.e. `../something`)?
> > 
> > 
> > > 
> > 
> > 
> > > Yes, i have tested it with numaflow repo, which has the same behaviour, as gitsync example-
> > 
> > 
> > > 
> > 
> > 
> > > Url: https://github.com/numaproj/numaflow.git Path: config/namespace-install
> > 
> > 
> > > 
> > 
> > 
> > > Here path contains the kustomization file which is referring resource as [../base](https://github.com/numaproj/numaflow/blob/main/config/namespace-install/kustomization.yaml#L5)
> > 
> > 
> > And did you try modifying something from `../base`?
> 
> No haven't this scenario, but I believe it should work because modifying anything from ../base will change commit history so the controller will pick it up and apply it.

Got it. Yeah, I see the condition is `if remoteRef.String() != lastCommitHash` so if any change occurred in the repository you'll do the comparison. Thanks.

xdevxy (commented):
Have you tried your change in the local cluster with kustomized manifests?

internal/git/gitsync_processor_test.go
```diff
@@ -685,9 +687,20 @@ func Test_watchRepo(t *testing.T) {
...
+		{
+			name: "Apply manifest from kustomize enabled repo",
+			gitSync: newGitSync(v1alpha1.RepositoryPath{
+				RepoUrl:        "https://github.com/numaproj/numaflow.git",
+				Path:           "config/namespace-install",
+				TargetRevision: "main",
+				Name:           "numaflow",
+			}),
+			hasErr: false,
+		},
```
juliev0, line 703: Maybe in https://github.com/numaproj-labs/numaplane-control-manifests.git or another repo that we control we could have all of the manifests we depend on for testing, rather than being dependent on this repository. I'll write an issue for that.

juliev0: https://github.com/numaproj-labs/numaplane/issues/101

juliev0 (approved):

chandankumar4:
> I guess for the test above you tried your change in the local cluster with kustomized manifests?

yes, I have tested it in local cluster

Issue #83, opened 2024-02-12
Update GitSync CRD to only have one RepositoryPath and one destination


xdevxy:
Based on the revised [design](https://docs.google.com/document/d/1nMP8QmhqsiiDDn-QSEuQr_1pA_mN5-0TxNPlaQPNlyQ/edit?usp=sharing) here, we would want to update the GitSync CRD accordingly.

PR #89, opened 2024-02-13
chore: Update GitSync CRD to only have one RepositoryPath and one destination

xdevxy:
Closes https://github.com/numaproj-labs/numaplane/issues/83


api/v1alpha1/gitsync_types.go
```diff
@@ -140,23 +140,16 @@ func init() {
...
-// ContainsClusterDestination determines if the cluster is in the list of Destinations
+// ContainsClusterDestination determines if the cluster matches the Destination
 func (gitSyncSpec *GitSyncSpec) ContainsClusterDestination(cluster string) bool {
-	for _, destination := range gitSyncSpec.Destinations {
-		if destination.Cluster == cluster {
-			return true
-		}
-	}
-	return false
+	return gitSyncSpec.Destination.Cluster == cluster
```
whynowy, line 145: Do we still need the `cluster`?

xdevxy: yeah, let's keep it just for extra safety to indicate this is really meant for this cluster.

whynowy: What I meant was do we still need the cluster field in the spec?

internal/git/gitsync_processor.go
```diff
@@ -137,21 +137,21 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1alpha1.GitSync
...
 
 		lastCommitHash = hash.String()
 	} else {
-		lastCommitHash = gitSync.Status.CommitStatus[repo.Name].Hash
+		lastCommitHash = gitSync.Status.CommitStatus.Hash
 	}
 	// no monitoring if targetRevision is a specific commit hash
 	if isCommitSHA(repo.TargetRevision) {
-		if val.Hash != hash.String() {
+		if gitSync.Status.CommitStatus.Hash != hash.String() {
```
juliev0, line 151: I know this was in here before, but I wonder if it's supposed to say `if lastCommitHash != hash.String()`? (for the case that there is nothing stored in the Status previously?)

juliev0:
Just remembered I was going to remove the memory leak from the gitSyncLocks, but I can write an issue and assign it to myself to do in a subsequent PR

api/v1alpha1/gitsync_types.go
xdevxy: Yeah, I was actually responding to that, :)

internal/git/gitsync_processor.go
xdevxy: Yes, you're right. It is more appropriate to use `lastCommitHash` here.

internal/git/gitsync_processor.go
```diff
@@ -412,10 +411,8 @@ func updateCommitStatus(
 			return err
 		}
 	}
-	if gitSync.Status.CommitStatus == nil {
-		gitSync.Status.CommitStatus = make(map[string]v1alpha1.CommitStatus)
-	}
-	gitSync.Status.CommitStatus[repo.Name] = commitStatus
+
+	gitSync.Status.CommitStatus = commitStatus
 	// It's Ok to fail here as upon errors the whole process will be retried
```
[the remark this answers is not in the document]
xdevxy: We do retry https://github.com/numaproj-labs/numaplane/blob/main/internal/git/gitsync_processor.go#L456-L461

internal/git/gitsync_processor.go
```diff
@@ -137,21 +137,21 @@ func watchRepo(ctx context.Context, r *git.Repository, gitSync *v1alpha1.GitSync
...
+		lastCommitHash = gitSync.Status.CommitStatus.Hash
 	}
 	// no monitoring if targetRevision is a specific commit hash
 	if isCommitSHA(repo.TargetRevision) {
-		if val.Hash != hash.String() {
+		if lastCommitHash != hash.String() {
 			logger.Errorw(
 				"synced commit hash doesn't match desired one",
-				"synced commit hash", val.Hash,
+				"synced commit hash", gitSync.Status.CommitStatus.Hash,
```
juliev0, line 154:
```suggestion
				"synced commit hash", lastCommitHash,
```

internal/git/gitsync_processor.go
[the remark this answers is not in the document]
juliev0: Since we are setting `lastCommitHash`, won't we skip retrying anything since it won't be new?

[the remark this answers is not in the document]
xdevxy: lastCommitHash  is only used within watchRepo, right? So I don't see it will lead to us skip anything.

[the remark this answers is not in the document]
juliev0: nice, you're right - sorry for the confusion

juliev0 (approved):

Issue #9, opened 2024-01-04
Authn/authz setup for accessing a git repository


xdevxy:
Numaplane ConfigMap should contain any Git Repository Prefix that has credentials for repositories nested within. For any given repository, there may or may not be a matching prefix in there. If there is no matching entry, then no credentials are needed for cloning that RepositoryPath.

For any entry in that ConfigMap, it may contain:
- no credentials
- SSHCredential only
- HTTPCredential only
- TLSCredential only (theoretically I think this should be possible even if it's never used in practice?)
- SSHCredential, HTTPCredential
- SSHCredential, TLSCredential
- HTTPCredential, TLSCredential
- SSHCredential, HTTPCredential, TLSCredential

Our RepositoryPath contains a protocol in it ("ssh" or "http" or "https" (or maybe another one but let's at least support these)). Based on that protocol, we need to see if there is an "HTTPCredential" or "SSHCredential". 

For "https" there should also be a TLSCredential. For "http" there should not. For "ssh" there could potentially be a TLSCredential.

chandankumar4:
i'll work on it :)

juliev0:
@xdevxy what are the details here? Basically we need to store github credentials in a Secret or something? In order for @shubhamdixit863 to test https://github.com/numaproj-labs/numaplane/issues/4, he will need to use some sort of github credentials anyway, right? But perhaps he won't need to store them in a Secret, and that can be the work of this Issue? Is that what you have in mind?

xdevxy:
Yeah @shubhamdixit863 can use a public repo for testing to start.

juliev0:
> Yeah @shubhamdixit863 can use a public repo for testing to start.

Oh, okay. No credentials required in that case I guess. 

So, for this feature the RepositoryPaths in the CRD can each include an optional Secret containing credentials, or something like that.

xdevxy:
Yeah, right, updating the issue's description to reflect that.

juliev0:
@shubhamdixit863 In the RepositoryPath you can include a Secret similar to [this](https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/redis_buffer_service.go#L50) one used by Numaflow RedisBufferService

@xdevxy which [authentication](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github) methods will we offer (initially)?

xdevxy:
[...]

PR #94, opened 2024-02-15
Added Authentication methods For git clone

shubhamdixit863:
Fixes https://github.com/numaproj-labs/numaplane/issues/9
### Modifications

[GetAuthMethod](https://github.com/numaproj-labs/numaplane/blob/d992ca903e9c9510185794b17c1643f7aeab077f/internal/shared/gitconfig/gitconfig.go#L80)   has been added which returns the  auth method as per the credentials from following [file](https://github.com/numaproj-labs/numaplane/blob/d992ca903e9c9510185794b17c1643f7aeab077f/config/samples/config.yaml)  .It is assumed that for a particular repository URL, one of the methods—HTTP, SSH, or TLS—has been set. If none is set then auth method would be nil and its assumed that its a public repository and can be downloaded without authentication. Moreover [FindCredsByUrl](https://github.com/numaproj-labs/numaplane/blob/53d31bfc76d37dd584f3c36b4c5f7e413a641e9f/internal/shared/git/git.go#L181) has been added which checks for the presence of credentials for github repository mentioned in [GitSync Crd](https://github.com/numaproj-labs/numaplane/blob/53d31bfc76d37dd584f3c36b4c5f7e413a641e9f/config/samples/numaplane.numaproj.io_v1alpha1_gitsync.yaml) by matching the prefix example if the repository is `https://github.com/numaproj-labs/numaplane-control-manifests.git` then configmap should contain credentials for the prefix `github.com/numaproj-labs` .

### Verification

Unit Tests
As of now only ssh and http cloning has been tested With ssh [method](https://github.com/numaproj-labs/numaplane/blob/d992ca903e9c9510185794b17c1643f7aeab077f/internal/git/gitsync_processor_test.go#L749) and [http](https://github.com/numaproj-labs/numaplane/blob/7d398242bcdf7fb9c317c05b77fe0f4ca2b80663/internal/git/gitsync_processor_test.go#L736)


config/manager/manager.yaml
```diff
@@ -70,10 +70,8 @@ spec:
         - /manager
         args:
         - --leader-elect
-        env:
-          - name: CLUSTER_NAME
-            value: CLUSTER_NAME_VALUE # Replace it with the cluster name where numaplane will get deployed.
         image: numaplane-controller:latest
+        imagePullPolicy: IfNotPresent
```
juliev0, line 74: accidental change?

internal/git/gitsync_processor.go
```diff
@@ -462,6 +496,48 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeCli
 	return processor, nil
 }
 
+func GetAuthMethod(ctx context.Context, repoUrl string, kubeClient kubernetes.Client, namespace string, repoCred *controllerconfig.GitCredential) (transport.AuthMethod, error) {
+	scheme, err := validations.GetTransportScheme(repoUrl)
```
juliev0, line 500: it's weird to call from here into `validations` code. Let's move this method to somewhere in "internal/shared" maybe so both places can get to it

internal/git/gitsync_processor.go
```diff
@@ -462,6 +496,48 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeCli
...
+	if err != nil {
+		return nil, err
+	}
+	var authMethod transport.AuthMethod
+	switch scheme {
+	case "ssh", "git+ssh":
```
juliev0, line 506: I didn't know about "git+ssh" - do you know if this is the "git" protocol mentioned [here](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols)? (Interestingly, that link there says there's no authentication for it so it's possible we don't go through this logic for that.) Have you found this used anywhere? Is there some documentation you used in the go-git reference that you can point me to for this?

shubhamdixit863: I found this repository -https://github.com/whilp/git-urls/blob/master/urls.go ,it has mentioned git+ssh transports

internal/git/gitsync_processor.go
```diff
@@ -443,9 +473,13 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeCli
 		channel:     channel,
 		clusterName: clusterName,
 	}
-
 	go func(repo *v1alpha1.RepositoryPath) {
-		r, err := cloneRepo(repo)
+		credentials := controllerconfig.GetConfigManagerInstance().GetConfig().RepoCredentials
+		method, err := GetAuthMethod(ctx, repo.RepoUrl, kubeClient, namespace, credentials[strings.ToLower(repo.RepoUrl)]) // viper converts to all lower
```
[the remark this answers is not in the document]
shubhamdixit863: GetAuthMethod basically is for putting up the authenication type whether ssh or http .For the other questions will the urls in crd and config would be mentioned without transport like http or ssh ?

juliev0:
> No
> 
> Fixes #TODO
> 
> ### Modifications
> Added Authentication methods For git clone
> 
> ### Verification
> Unit Tests

Hi @shubhamdixit863. In general, it will be helpful for me if you can be specific about what cases your Unit Tests test in your PR description for "Verification". This will help me to be able to quickly see what you tested so I can think about what wasn't tested. Thank you!

shubhamdixit863:
> > No
> > Fixes #TODO
> > ### Modifications
> > Added Authentication methods For git clone
> > ### Verification
> > Unit Tests
> 
> Hi @shubhamdixit863. In general, it will be helpful for me if you can be specific about what cases your Unit Tests test in your PR description for "Verification". This will help me to be able to quickly see what you tested so I can think about what wasn't tested. Thank you!

Noted

Makefile
```diff
@@ -75,6 +75,12 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -v ./... -coverprofile cover.out
 
+
+.PHONY: test-with-race
+test-with-race: manifests generate fmt vet envtest ## Run tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -race ./... -coverprofile cover.out
```
juliev0, line 81: Please set this the same way as the ConfigMap [PR](https://github.com/numaproj-labs/numaplane/pull/80/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52)

config/samples/config.yaml
```diff
@@ -0,0 +1,26 @@
...
+    HTTPCredential:
+      Username: "exampleUser"
+      Password:
+        name: "http-creds"
+        key: "password"
+  "git@github.com:rustyTest/testprivateRepo.git":
+    SSHCredential:
+      SSHKey:
+        name: "ssh-creds"
+        key: "sshKey"
```
juliev0, line 14:
```suggestion
  "github.com/rustyTest/testprivateRepo":
    HTTPCredential:
      Username: "exampleUser"
      Password:
        name: "http-creds"
        key: "password"
    SSHCredential:
      SSHKey:
        name: "ssh-creds"
        key: "sshKey"
```

juliev0: the prefix actually shouldn't take into account the protocol ("ssh" vs "https")

juliev0: (so you'll probably need to change your logic if it is)

internal/git/gitsync_processor.go
```diff
@@ -74,9 +77,15 @@ type MetaData struct {
 	Namespace string `yaml:"namespace"`
 }
 
-func cloneRepo(repo *v1alpha1.RepositoryPath) (*git.Repository, error) {
+func cloneRepo(repo *v1alpha1.RepositoryPath, authMethod transport.AuthMethod) (*git.Repository, error) {
+	// Add any certificates if its required
```
juliev0, line 81:
```suggestion
	// Add any certificates if it's required
```

internal/shared/gitconfig/gitconfig.go
```diff
@@ -0,0 +1,131 @@
...
+
+func IsValidKubernetesNamespace(name string) bool {
+	// All namespace names must be valid RFC 1123 DNS labels.
+	errs := validation.IsDNS1123Label(name)
+	reservedNamesRegex := regexp.MustCompile(`^(kubernetes-|kube-)`)
+	if len(errs) == 0 && !reservedNamesRegex.MatchString(name) {
+		return true
+	}
+	return false
+}
```
juliev0, line 88: It seems odd for this method to be in a file called gitconfig.go.

Also, do any of the other functions in here have anything to do with "config" per se? Or could they really live in a file called git.go?

config/samples/config.yaml
```diff
@@ -0,0 +1,26 @@
+clusterName: "example-cluster"
+timeIntervalSec: 60
+repoCredentials:
+  "https://github.com/rustyTest/testprivateRepo":
```
whynowy, line 4: This kind of configuration looks weird to me. 

1. Was it because the k8s `secretKeySelecotr` can not be used to directly with viper, we end up with writing our own?
2. `SSHCredential`, `SSHKey`... do not follow the same pattern as others (lower case first letter).
3. A key like `"git@github.com:rustyTest/testprivateRepo.git"` is tricky, what is the reason we do not user an array?

internal/git/gitsync_processor.go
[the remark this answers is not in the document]
juliev0: > will the urls in crd and config would be mentioned without transport like http or ssh ?

yes, the urls shouldn't include the transport (which I referred to as "protocol" elsewhere)

internal/git/gitsync_processor.go
juliev0: thanks for noticing that

internal/shared/gitconfig/gitconfig.go
```diff
@@ -0,0 +1,131 @@
...
+	)
+)
+
+// Checks for valid remote repository
+
+type TransportSet struct {
+	Transports map[string]struct{}
+}
+
+// NewTransportSet returns a TransportSet with the items keys mapped
```
juliev0, line 41:
```suggestion
// NewTransportSet returns a TransportSet with the items' keys mapped
```

config/samples/config.yaml
juliev0: > 3\. A key like `"git@github.com:rustyTest/testprivateRepo.git"` is tricky, what is the reason we do not use an array?
It should be:
```
repoCredentials: 
   "github.com/rustyTest/testprivateRepo":
        SSHCredential:
          ...
        HTTPCredential:
          ....
```

As far as array, I made the assumption that a user should not need more than one ssh credential or http credential. Do you disagree?

whynowy: What I meant was to make the value of `repoCredentials` an array, so that we do not need to use a key like `github.com/rustyTest/testprivateRepo`, instead, to use something like `url: github.com/rustyTest/testprivateRepo`.

xdevxy: > What I meant was to make the value of repoCredentials an array, so that we do not need to use a key like github.com/rustyTest/testprivateRepo, instead, to use something like url: github.com/rustyTest/testprivateRepo.

Yeah, that makes sense.

juliev0: sure

internal/shared/gitconfig/gitconfig.go
shubhamdixit863: So should i make the file name shared/git/git.go ?

config/samples/config.yaml
shubhamdixit863: go-git throws error repo not found ,if used plane repo -`github.com/rustyTest/testprivateRepo.git` see this -https://github.com/numaproj-labs/numaplane/blob/7d398242bcdf7fb9c317c05b77fe0f4ca2b80663/internal/git/gitsync_processor_test.go#L770

internal/controller/config/config_test.go
```diff
@@ -20,38 +20,60 @@ func TestLoadConfigMatchValues(t *testing.T) {
...
-	assert.Equal(t, "tls.key", config.RepoCredentials["key3"].TLS.KeySecret.Key, "KeySecret Key for TLS does not match")
+	// Function to find the credentials for a given URL
+	findCredsByUrl := func(url string) *GitCredential {
+		for _, cred := range config.RepoCredentials {
+			if cred.URL == url {
+				log.Println("printed", cred.URL)
+				return cred.Credential
+			}
+		}
+		return nil
```
juliev0, line 44: nice

internal/shared/gitconfig/gitconfig.go
juliev0: `shared/git/git.go` seems good to me

internal/controller/gitsync_controller.go
```diff
@@ -299,7 +295,7 @@ func (r *GitSyncReconciler) validate(gitSync *apiv1.GitSync) error {
 	if len(destination.Cluster) == 0 {
 		return fmt.Errorf("cluster name cannot be empty")
 	}
-	if !validations.IsValidKubernetesNamespace(destination.Namespace) {
+	if !gitconfig.IsValidKubernetesNamespace(destination.Namespace) {
```
juliev0, line 298: make sure this method is moved somewhere kubernetes related, and not git

internal/controller/config/config.go
```diff
@@ -116,6 +108,19 @@ func (cm *ConfigManager) LoadConfig(onErrorReloading func(error), configPath str
 	return nil
 }
 
+// LoadConfigFromBuffer is  Specifically for tests
```
juliev0, line 111:
```suggestion
// LoadConfigFromBuffer is specifically for tests
```

juliev0: Is this function an alternative to testing `LoadConfig()`? It's a little odd to have a different function being tested than the real one. We won't catch if that changes. Is it necessary?

internal/git/gitsync_processor.go
```diff
@@ -601,7 +584,18 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeCli
...
+		}
+
+		gitCredentials := gitshared.FindCredByUrl(repo.RepoUrl, globalConfig)
+		method, err := gitshared.GetAuthMethod(ctx, gitCredentials, kubeClient, namespace)
+		if err != nil {
+			logger.Errorw("error getting  the  auth method", "err", err)
+		}
+
+		r, err := cloneRepo(ctx, localRepoPath, repo, method)
 		if err != nil {
```
juliev0, line 599: FYI, I have added [this](https://github.com/numaproj-labs/numaplane/issues/108) Issue to ensure we use the new Credentials in the case that Credentials change in the ConfigMap while we are reading from a repo (if we get an error). This is something that may make more sense to do after the Worker Queue refactor.

internal/shared/git/git.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			return nil, err
+		}
+		authMethod, err = ssh.NewPublicKeys("git", secret.Data["sshKey"], "")
+		if err != nil {
+			return nil, err
+		}
+
+	case repoCred.TLS != nil:
+		// TODO :this needs to be implemented
```
juliev0, line 104: are you thinking of doing it in a subsequent issue? If so, can you create an Issue for it?

internal/controller/config/config.go
shubhamdixit863: I used it to loadconfig in gitcontroller_tests ,I will see if this can be removed and we can use existing LoadConfig

internal/git/gitsync_processor.go
shubhamdixit863: So Do I need to change anything in here or should I leave it as it is for now @juliev0

internal/git/gitsync_processor.go
```diff
@@ -601,7 +584,18 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeCli
 	}
 	localRepoPath := getLocalRepoPath(gitSync.GetName())
 	go func(repo *v1alpha1.RepositoryPath) {
-		r, err := cloneRepo(ctx, localRepoPath, repo)
+		globalConfig, err := controllerconfig.GetConfigManagerInstance().GetConfig()
+		if err != nil {
+			logger.Errorw("error getting  the  global config", "err", err)
+		}
+
+		gitCredentials := gitshared.FindCredByUrl(repo.RepoUrl, globalConfig)
```
juliev0, line 592: It should be allowed for a given repository that there are no credentials and so nothing in the ConfigMap. Need to account for that case.

internal/kubernetes/interface.go
```diff
@@ -24,6 +25,9 @@ type Client interface {
 	// Update the resource of kubernetes cluster.
 	Update(ctx context.Context, obj k8sClient.Object, opts ...k8sClient.UpdateOption) error
 
+	// GetSecret Gets Secret For the Kubernetes
```
juliev0, line 28:
```suggestion
	// GetSecret Gets a Kubernetes Secret
```

internal/shared/git/git.go
```diff
@@ -0,0 +1,121 @@
...
+		"rsync",
+		"file",
+	)
+)
+
+// Checks for valid remote repository
+
+type TransportSet map[string]struct{}
+
+// NewTransportSet returns a TransportSet with the items keys mapped
```
juliev0, line 40:
```suggestion
// NewTransportSet returns a TransportSet with the items' keys mapped
```

internal/shared/git/git.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func IsValidKubernetesNamespace(name string) bool {
+	// All namespace names must be valid RFC 1123 DNS labels.
+	errs := validation.IsDNS1123Label(name)
+	reservedNamesRegex := regexp.MustCompile(`^(kubernetes-|kube-)`)
+	if len(errs) == 0 && !reservedNamesRegex.MatchString(name) {
+		return true
+	}
+	return false
```
juliev0, line 73: make sure to move this out of git.go

internal/git/gitsync_processor.go
juliev0: It should also be allowed that this prefix is in the ConfigMap but there are no credentials with it. Please make sure that you account for and test both of these cases.

internal/git/gitsync_processor.go
juliev0: nothing to do now, thanks

config/samples/config.yaml
```diff
@@ -1,26 +1,69 @@
...
-        key: "ca.crt"
-      certSecret:
-        name: "cert-secret"
-        key: "tls.crt"
-      KeySecret:
-        name: "key-secret"
-        key: "tls.key"
+  - url: "github.com/rustyTest/testprivateRepo"
+    credential:
+      HTTPCredential:
```
whynowy, line 6: Can we update all of these `json` fields to following?

`httpCredential`
`username`
`password`
`sshCredential`
`sshKey`
`tls`
`caCertSecret`
`keySecret`

Follow the practice of Kubernetes.

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/autoscaling/v1/types.go#L36
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/networking/v1/types.go#L301

shubhamdixit863: Addressed

shubhamdixit863: Yeah pushed just now @whynowy

juliev0:
> It is assumed that for a particular repository URL, one of the methods—HTTP, SSH, or TLS—has been set. 

So, I think TLS can be used in conjunction with HTTP (in which case it's basically HTTPS) or it could be used with SSH.
Unless I'm mistaken, I think all of the possible credential permutations for a given connection would include:
- none
- http password(includes access token)
- ssh key
- http password+tls
- ssh key+tls

config/samples/config.yaml
```diff
@@ -1,26 +1,69 @@
...
+        insecureSkipVerify: true
+        caCertSecret:
+          name: "ca-cert-secret"
+          key: "ca.crt"
+        certSecret:
+          name: "cert-secret"
+          key: "tls.crt"
+        keySecret:
+          name: "key-secret"
+          key: "tls.key"
```
juliev0, line 25: As far as I understand, I believe we would either use `insecureSkipVerify` or define all of those other fields, but not both.

config/samples/config.yaml
```diff
@@ -1,26 +1,69 @@
...
+  - url: "git@github.com:numaproj"
```
juliev0, line 26:
```suggestion
  - url: "github.com:numaproj"
```

juliev0: should not contain the protocol

config/samples/config.yaml
```diff
@@ -1,26 +1,69 @@
 clusterName: "example-cluster"
```
juliev0, line 1: I just realized this is some free form yaml as opposed to a Kubernetes ConfigMap. I figured that this directory would hold actual full Kubernetes resource definitions, just like the GitSync that lives in this directory. Any reason it doesn't?

Also, do we need 2 files (I see you have `testConfig.yaml` as well) or can we just have 1?

internal/shared/git/git.go
```diff
@@ -0,0 +1,200 @@
...
+	_, ok := t[transport] // Directly checking the key in the map
+	return ok
+}
+
+func CheckGitURL(gitURL string) bool {
+	u, err := url.Parse(gitURL)
+	if err == nil && !Transports.Valid(u.Scheme) {
+		return false
+	}
+	log.Println(u)
```
juliev0, line 64: make sure you remove any Println statements you added for testing

internal/shared/git/git.go
```diff
@@ -0,0 +1,200 @@
...
+		secret, err := kubeClient.GetSecret(ctx, namespace, key)
+		if err != nil {
+			return nil, err
+		}
+		authMethod, err = ssh.NewPublicKeys("git", secret.Data["sshKey"], "")
+		if err != nil {
+			return nil, err
+		}
+
+	case repoCred.TLS != nil:
```
juliev0, line 170: The way this should work is that you take whatever protocol the RepositoryPath is (http or ssh) and then see if there's a corresponding HTTPCredential or SSHCredential depending on that protocol. Then there may or may not be a TLSCredential in addition.

config/samples/config.yaml
shubhamdixit863: The config.yaml file will be mounted as a volume because it is parsed through Viper and will be loaded from /etc/numaplane inside the pod. The other config map you mentioned is located in the manager folder, as seen [here](https://github.com/numaproj-labs/numaplane/blob/feat/repoAuth/config/manager/config.yaml).

internal/shared/git/git.go
shubhamdixit863: Ok ,Acknowledged .I am addressing that in another PR

config/samples/config.yaml
juliev0: > The testConfig.yaml I have created is specifically for controller tests, as it contains different values for the configuration to keep things isolated. Moreover, due to frequent changes in the gitsync processor , controller tests often fail. To avoid such occurrences, I have created a separate testConfig.yaml. Should I make it one ?

I think maybe if there's a file that's for a particular test, we should colocate it with the test in a `testData` directory.

config/samples/config.yaml
shubhamdixit863: Ok ,Acknowledged .I am addressing that in another PR

config/samples/config.yaml
shubhamdixit863: Noted ,not needed as of now as I have merged them.

config/samples/config.yaml
juliev0: I know it's challenging having multiple PRs at the same time, but it will make sure we don't lose sight of things if we do things within the PR where the issue is caught. For example, your other PR may not end up getting merged in the end for whatever reason, so it's possible the issue will get forgotten.

juliev0:
Can you please expand your "Verification" section of your Description? More details will help me to know what was and wasn't tested without needing to go dig through the tests themselves. For example, I want to know that you made sure it works with no credentials set, etc. Thanks.

internal/shared/git/git.go
juliev0: In that case we should close this PR. Because this is wrong for this PR.

xdevxy (approved):

shubhamdixit863:
@juliev0 @whynowy Our initial plan for this [file](https://github.com/numaproj-labs/numaplane/pull/94/commits/1463f6260d9d47243c1a6788a8dcb5fc7139aa7b) was to include only the organization part or username, for example, "github.com/numaproj-labs." Here, "github.com" serves as the domain of the hosting service, indicating that the repository is hosted on GitHub, and "numaproj-labs" represents the username or organization name under which the repository is located. Given this context, are we now transitioning to include the complete repository name, such as "github.com/numaproj-labs/numaplane-test"?  . If that's the case then  should we  do full url match instead of [prefix](https://github.com/numaproj-labs/numaplane/blob/ea588e21fa2de632a580b246220e7627fe982603/internal/shared/git/git.go#L178) match ?

shubhamdixit863:
> > @juliev0 @whynowy Our initial plan for this [file](https://github.com/numaproj-labs/numaplane/pull/94/commits/1463f6260d9d47243c1a6788a8dcb5fc7139aa7b) was to include only the organization part or username, for example, "github.com/numaproj-labs." Here, "github.com" serves as the domain of the hosting service, indicating that the repository is hosted on GitHub, and "numaproj-labs" represents the username or organization name under which the repository is located. Given this context, are we now transitioning to include the complete repository name, such as "github.com/numaproj-labs/numaplane-test"? . If that's the case then should we do full url match instead of [prefix](https://github.com/numaproj-labs/numaplane/blob/ea588e21fa2de632a580b246220e7627fe982603/internal/shared/git/git.go#L178) match ?
> 
> I see. I didn't know that. Will update it.

yes ,I have made the test cases as per that . If thats good with you ,I will update the files back to original .

shubhamdixit863:
Closed  as opening new one,

PR #97, opened 2024-02-15
Mark `SyncTime` and `Hash` omitempty to avoid validation when empty

### LINKED ISSUE #96: Empty CommitStatus trigger validation error
When apply a GitSync CR now, we saw the following error.

`{"level":"error","ts":"2024-02-15T23:38:06.917335+05:30","logger":"numaplane","caller":"controller/gitsync_controller.go:129","msg":"Error Updating GitSync Status","err":"GitSync.numaplane.numaproj.io \"gitsync-sample\" is invalid: status.commitStatus.syncTime: Required value","GitSync":"numaplane-system/gitsync-sample","stacktrace":"github.com/numaproj-labs/numaplane/internal/controller.(*GitSyncReconciler).Reconcile\n\t/Users/ckumar14/go/src/github.com/numaproj-labs/numaplane/internal/controller/gitsync_controller.go:129\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\t/Users/ckumar14/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:119\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/Users/ckumar14/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:316\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/ckumar14/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/Users/ckumar14/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:227"}`

We should mark CommitStatus field as omitempty to avoid validation when empty.

xdevxy:
Fixes #96 

### Modifications

Mark  `SyncTime` and `Hash` in `CommitStatus` omitempty

### Verification

In local cluster,
```
make deploy
```


api/v1alpha1/gitsync_types.go
```diff
@@ -98,13 +98,13 @@ type Destination struct {
...
 	// Hash of the git commit
-	Hash string `json:"hash"`
+	Hash string `json:"hash,omitempty"`
 
 	// Synced indicates if the sync went through
 	Synced bool `json:"synced,omitempty"`
 
 	// SyncTime represents the last time that we attempted to sync this commit (whether or not it succeeded)
-	SyncTime metav1.Time `json:"syncTime"`
+	SyncTime metav1.Time `json:"syncTime,omitempty"`
```
juliev0, line 107: Thanks for fixing but trying to understand what happened. I see our `CommitStatus` [field](https://github.com/numaproj-labs/numaplane/pull/97/files#diff-d43361aee7a4a6b3365a5445375072d8587995c112f648dd0eb8132549f1d391R67) itself is already "omitempty". So, do we assume that one got pushed but with these fields not set?

xdevxy: yeah I think it is related to we take out the mapping to repo from `CommitStatus`. Now we only have one CommitStatus field, which may check inside what is the value there.

juliev0: oh, in that case I wonder if the `CommitStatus` should be a pointer? I wonder if that would take care of it? And then maybe we could still have fields in here that are mandatory if you do have a `CommitStatus`

chandankumar4: I was just going though how k8s handle this, so they handle the same ways as @juliev0  mentioned, like we have `CommitStatus` as pointer and inside that we have each field as omitempty. As example [StatusDetails](https://github.com/kubernetes/kubernetes/blob/91ee30074bee617d52fc24dc85132fe948aa5153/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L776) is used as pointer [here](https://github.com/kubernetes/kubernetes/blob/91ee30074bee617d52fc24dc85132fe948aa5153/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L764) with `omitempty`

xdevxy: Sounds good!

xdevxy: @chandankumar4 Although in your example all fields in `StatusDetails` is `omitempty`.

api/v1alpha1/gitsync_types.go
```diff
@@ -101,9 +101,9 @@ type CommitStatus struct {
 	Hash string `json:"hash"`
 
 	// Synced indicates if the sync went through
-	Synced bool `json:"synced,omitempty"`
+	Synced bool `json:"synced"`
 
-	// SyncTime represents the last time that we attempted to sync this commit (whether or not it succeeded)
+	// SyncTime represents the last time that we attempted to sync this commit (whether it succeeded)
```
juliev0, line 106: oh, do you mean that we won't set the time if it doesn't succeed? (my original thought is that we would)

juliev0, line 106:
```suggestion
	// SyncTime represents the last time that we attempted to sync this commit (regardless of whether it succeeded)
```

juliev0: oh, I have to admit I don't like the IDE's suggestion. :) If the IDE doesn't like the original one, I recommended another one below.

juliev0 (approved):

PR #111, opened 2024-02-22
chore: Ability to run in local cluster without pushing images

whynowy:
`make deploy` to run the controller in a local cluster without pushing images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,16 @@
+# Setting SHELL to bash allows bash commands to be executed by recipes.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
 
 # Image URL to use all building/pushing image targets
 IMG ?= numaplane-controller
 VERSION ?= latest
+BASE_VERSION := latest
 # Default cluster name where numaplane get deployed, update it as needed.
 CLUSTER_NAME ?= "staging-usw2-k8s"
-#TODO: add back once ready to push to quay.io
-#IMAGE_NAMESPACE ?= quay.io/numaproj
-#IMAGE_FULL_PATH ?= $(IMAGE_NAMESPACE)/$(IMG):$(VERSION)
-IMAGE_FULL_PATH ?= $(IMG):$(VERSION)
+IMAGE_NAMESPACE ?= quay.io/numaproj
+IMAGE_FULL_PATH ?= $(IMAGE_NAMESPACE)/$(IMG):$(VERSION)
 
 GCFLAGS="all=-N -l"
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,13 +1,16 @@
+# Setting SHELL to bash allows bash commands to be executed by recipes.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
 
 # Image URL to use all building/pushing image targets
 IMG ?= numaplane-controller
 VERSION ?= latest
+BASE_VERSION := latest
```
whynowy, line 9: Will be useful for future release scripts.

config/manager/config.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: numaplane-controller-config
+  name: controller-config
   namespace: numaplane-system
 data:
   config.yaml: |
-    clusterName: "CLUSTER_NAME_VALUE" # Replace with the cluster name where numaplane will get deployed.
+    clusterName: CLUSTER_NAME_VALUE # Replace with the cluster name where numaplane will get deployed.
```
whynowy, line 8: The extra `"` was making trouble.

config/manager/manager.yaml
```diff
@@ -103,4 +103,4 @@ spec:
       volumes:
         - name: config-volume
           configMap:
-            name: numaplane-controller-config
+            name: controller-config
```
whynowy, line 106: Having the prefix made the configmap name `numaplane-numaplane-controller-config`.

Makefile
```diff
@@ -1,13 +1,16 @@
...
 # Default cluster name where numaplane get deployed, update it as needed.
-CLUSTER_NAME ?= "staging-usw2-k8s"
-#TODO: add back once ready to push to quay.io
-#IMAGE_NAMESPACE ?= quay.io/numaproj
-#IMAGE_FULL_PATH ?= $(IMAGE_NAMESPACE)/$(IMG):$(VERSION)
-IMAGE_FULL_PATH ?= $(IMG):$(VERSION)
+CLUSTER_NAME ?= staging-usw2-k8s
```
whynowy, line 11: The `"` was making trouble.

Makefile
```diff
@@ -146,9 +160,8 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 	$(KUSTOMIZE) build config/crd | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: deploy
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	cd config/manager && $(KUSTOMIZE) edit set image numaplane-controller=${IMAGE_FULL_PATH}
```
whynowy, line 150: Should not use `kustomize edit` here, it updates `kustomization.yaml`.

Makefile
```diff
@@ -1,13 +1,16 @@
+# Setting SHELL to bash allows bash commands to be executed by recipes.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
 
 # Image URL to use all building/pushing image targets
 IMG ?= numaplane-controller
 VERSION ?= latest
```
xdevxy, line 8: What is the difference between `VERSION` and `BASE_VERSION`?

xdevxy (commented):
Did you also try `make deploy` with image on remote registry is still working?

whynowy:
> Did you also try `make deploy` with image on remote registry is still working?

This command is supposed to be used for development in local cluster only, in that case, we should not ask the developers to use a remote registry. If we want to deploy a remote image to a cluster (no matter if it's a local cluster or a remote one), for example, assume we have an official released version, and we want to verify if that release is good, then we should follow the released instruction to install (e.g. kubectl apply -f <released-manifests-url>).

Makefile
juliev0: would it make sense to add a comment for it then?

whynowy: Added a short comment for that.

xdevxy:
> > Did you also try `make deploy` with image on remote registry is still working?
> 
> This command is supposed to be used for development in local cluster only, in that case, we should not ask the developers to use a remote registry. If we want to deploy a remote image to a cluster (no matter if it's a local cluster or a remote one), for example, assume we have an official released version, and we want to verify if that release is good, then we should follow the released instruction to install (e.g. kubectl apply -f ).

Got it, makes sense, thanks!

xdevxy (approved):

Issue #109, opened 2024-02-21
Implement TLS Authentication for Git Clone


shubhamdixit863:
# Summary
Implementing TLS authentication should ensure that all Git clone operations are performed over a secure channel. The implementation should verify the server's TLS certificate and establish a secure connection before proceeding with the clone operation.

What change needs making?
Add TLS authentication to [GetAutMethod](https://github.com/numaproj-labs/numaplane/blob/d992ca903e9c9510185794b17c1643f7aeab077f/internal/shared/gitconfig/gitconfig.go#L80)
Related issue -https://github.com/numaproj-labs/numaplane/issues/9
# Use Cases

When would you use this?

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

xdevxy:
Closed by #118

PR #118, opened 2024-02-26
Authentication for git clone

shubhamdixit863:
Yes 

Fixes #TODO
https://github.com/numaproj-labs/numaplane/issues/9 -Authentication setup for cloning
https://github.com/numaproj-labs/numaplane/issues/109 - TLS implementation 

### Modifications

In this PR, different authentication mechanisms for various Git URLs (HTTP, SSH, HTTPS) have been checked.

**Flow:**

While cloning a repository, `[RepoCredentials](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/controller/config/config.go#L39)` are obtained from the Global Config, as `RepoCredentials` is an array of `[RepoCredential](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/controller/config/config.go#L42)`.

[`FindCredByUrl`](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/shared/git/git.go#L217) is used to get the credential for the Git repository.

Once the credentials are fetched, [`GetRepoCloneOptions`](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/shared/git/git.go#L144) is called to return the appropriate clone option that contains the `AuthMethod`.

After that,[ `cloneRepo`](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/git/gitsync_processor.go#L98) is called by passing the required `cloneOptions` obtained.

### Verification

Multiple unit and integration tests have been conducted to check the expected behavior.

[Docker Test](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/git/gitsync_processor_test.go#L115) has been used to create a Docker container that has Git installed behind the Apache server with HTTP and HTTPS. The running Docker container has also  been configured to use SSH to clone the repository.

[`TestGitCloneRepoHTTPSWithLocalGitServer`](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/git/gitsync_processor_test.go#L1045) tests HTTPS Git cloning over a Docker Git server.

[`TestGitCloneRepoHTTPWithLocalGitServer`](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/git/gitsync_processor_test.go#L1006) tests HTTP Git cloning over a Docker Git server.

[`TestGitCloneRepoSshLocalGitServer`](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/git/gitsync_processor_test.go#L963) tests SSH Git cloning over a Docker Git server.

Additionally, tests with a GitHub URL have been conducted over HTTP using a passkey obtained from github, and over SSH using a private-public key pair.

[`TestGitCloneRepoSsh`](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/git/gitsync_processor_test.go#L924) - Git clone with SSH over a public-private key pair.

[`TestGitCloneRepoHTTP`](https://github.com/numaproj-labs/numaplane/blob/feat/gitAuthentication/internal/git/gitsync_processor_test.go#L882) - Git clone with HTTP using a passkey.


internal/shared/git/git.go
```diff
@@ -0,0 +1,237 @@
...
+		authMethod := &gitHttp.BasicAuth{
+			Username: username,
+			Password: password,
+		}
+		cloneOptions = &git.CloneOptions{
+			URL:             endpoint.String(),
+			Auth:            authMethod,
+			InsecureSkipTLS: repoCred.TLS.InsecureSkipVerify,
+		}
+		return cloneOptions, nil
```
juliev0, line 169: I believe TLS can be combined with SSH. I think TLS is an orthogonal concept to HTTP vs SSH. So, I am thinking if you can essentially remove the first case statement, and have the latter 2 set `cloneOptions`. Then you can take what you have, and if TLS is also set then set TLS settings in addition.

juliev0: Oh wait. Also, we need to be using the protocol defined by the RepositoryPath. Here's what I imagine in the ConfigMap:
```
repoCredentials:
  - url: "github.com/numaproj-labs"
    httpCredential:
      ...
    sshCredential:
      ...
    tlsCredential:
      ...
```
(Note that all of those sections are optional.)
When we clone, our RepositoryPath will be something like "https://github.com/numaproj/numaflow.git". We need to take that RepositoryPath and say "okay, it starts with https", so let me only look to see whether there's an HTTPCredential. Then, in addition I can look for a TLSCredential.

internal/controller/config/config.go
```diff
@@ -57,7 +57,7 @@ type SSHCredential struct {
 
 type TLS struct {
 	InsecureSkipVerify bool              `json:"insecureSkipVerify"`
-	CACertSecret       SecretKeySelector `json:"CACertSecret"`
+	CACertSecret       SecretKeySelector `json:"caCertSecret"`
 	CertSecret         SecretKeySelector `json:"certSecret"`
 	KeySecret          SecretKeySelector `json:"keySecret"`
 }
```
juliev0, line 63: Sorry, I was mistaken about incorporating all of those fields. They seem to be related to the client authenticating itself to the server. How about we just include `InsecureSkipVerify` field?

type TLS struct {	type TLS struct {
	InsecureSkipVerify bool              `json:"insecureSkipVerify"`		InsecureSkipVerify bool              `json:"insecureSkipVerify"`
}

shubhamdixit863: Yeah addressed.

internal/shared/git/git.go
shubhamdixit863: I have made the method switch as per the scheme of git url

internal/git/gitsync_processor.go
```diff
@@ -587,7 +580,7 @@ func getLatestCommitHash(repo *git.Repository, refName string) (*plumbing.Hash,
 // The function clones the Git repository specified in the GitSync object and starts a goroutine to watch the repository for changes.
 // The goroutine will continuously monitor the repository and apply any changes to the Kubernetes cluster.
 // If any error occurs during the repository watching process, it will be logged and the commit status will be updated with the error.
-func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeClient kubernetes.Client, clusterName string, repoCred map[string]*controllerconfig.GitCredential) (*GitSyncProcessor, error) {
+func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeClient kubernetes.Client, clusterName string, repoCred []controllerconfig.RepoCredential) (*GitSyncProcessor, error) {
```
juliev0, line 583: I don't believe you're using `repoCred`, right? Can you remove the parameter?

juliev0:
Did you verify it works for these cases?:

- prefix is in the ConfigMap but there are no credentials
- prefix not in the ConfigMap at all

internal/shared/git/git.go
```diff
@@ -0,0 +1,228 @@
...
+// FindCredByUrl searches for GitCredential by the specified URL within the provided GlobalConfig.
+// It returns the matching GitCredential if the specified URL starts with the URL of any RepoCredentials, otherwise returns nil.
+func FindCredByUrl(url string, config controllerconfig.GlobalConfig) *controllerconfig.RepoCredential {
+	normalizedUrl := NormalizeGitUrl(url)
+	for _, cred := range config.RepoCredentials {
+		if strings.HasPrefix(normalizedUrl, NormalizeGitUrl(cred.URL)) {
+			return &cred
+		}
+	}
+	return nil
```
juliev0, line 216: This function looks good

internal/git/gitsync_processor.go
```diff
@@ -601,7 +594,17 @@ func NewGitSyncProcessor(ctx context.Context, gitSync *v1alpha1.GitSync, kubeCli
 	}
 	localRepoPath := getLocalRepoPath(gitSync.GetName())
 	go func(repo *v1alpha1.RepositoryPath) {
-		r, err := cloneRepo(ctx, localRepoPath, repo)
+		globalConfig, err := controllerconfig.GetConfigManagerInstance().GetConfig()
+		if err != nil {
+			logger.Errorw("error getting  the  global config", "err", err)
+		}
+		gitCredentials := gitshared.FindCredByUrl(repo.RepoUrl, globalConfig)
+		cloneOptions, err := gitshared.GetRepoCloneOptions(ctx, gitCredentials, kubeClient, namespace, repo)
```
juliev0, line 602: If prefix is not found, `gitCredentials` will be nil and `GetRepoCloneOptions()` doesn't seem to be able to handle that.

internal/shared/git/git.go
```diff
@@ -0,0 +1,228 @@
...
+		rawquery = m[4]
+	}
+	return &url.URL{
+		Scheme:   "ssh",
+		User:     userinfo,
+		Host:     m[2],
+		Path:     m[3],
+		RawQuery: rawquery,
+	}, nil
+}
```
juliev0, line 130: Just out of curiosity, I'm assuming this was grabbed from somewhere? maybe in go-git or elsewhere? were you not able to reference the gomodule itself and had to copy it?

.github/workflows/ci.yaml
```diff
@@ -10,12 +10,19 @@ jobs:
   unit-tests:
     name: Unit Tests
     runs-on: ubuntu-latest
+    services:
```
whynowy, line 13: @dpadhiar - please check the solution in Argo CD, and come up with proper approach for us.

For now we can use this.

config/samples/config.yaml
```diff
@@ -1,26 +1,34 @@
-clusterName: "example-cluster"
+clusterName: "staging-usw2-k8s"
```
whynowy, line 1: This is very tricky to put the file in the samples directory, if it's only used for test cases, move it there or use a const string in the test cases.

internal/shared/k8/k8.go
```diff
@@ -0,0 +1,17 @@
+package k8
```
whynowy, line 1: This package name is tricky.

internal/shared/git/git.go
shubhamdixit863: Yes ,I have mentioned in comment too ,https://github.com/numaproj-labs/numaplane/blob/182c970e4a9e839264d03ab9406033bf5772bd71/internal/shared/git/git.go#L21 . Initially i thought of referencing as go modules but the library was not maintained so i thought to copy the code there weren't many function though - o[riginal source](https://github.com/whilp/git-urls)

internal/git/gitsync_processor.go
shubhamdixit863: Ok i will put few more tests for them .

config/samples/config.yaml
juliev0: yeah, for tests let's collocate this type of thing with the test itself.

@whynowy what's your impression for what should go in `samples` directory? For one, I think everything should be an actual Kubernetes spec, which this isn't. Out of curiosity, would it be appropriate to put a Numaplane ConfigMap example in here (since it is part of Numaplane as opposed an object that is reconciled by Numaplane)?

config/samples/config.yaml
```diff
@@ -1,26 +1,34 @@
...
+      caCertSecret:
         name: "ca-cert-secret"
         key: "ca.crt"
       certSecret:
         name: "cert-secret"
         key: "tls.crt"
-      KeySecret:
+      keySecret:
         name: "key-secret"
         key: "tls.key"
```
juliev0, line 19: Since we removed these tls fields, let's remove them here too.

internal/shared/k8/k8.go
juliev0: can you clarify what you mean @whynowy ? Would you prefer "kubernetes"?

whynowy: `k8` is not a valid/meaningful name.

internal/shared/git/git.go
```diff
@@ -0,0 +1,251 @@
...
+		return "", err
+	}
+	scheme := parsedUrl.Scheme
+	return scheme, nil
+}
+
+// GetRepoCloneOptions creates git.CloneOptions for cloning a repo with HTTP, SSH, or TLS credentials from Kubernetes secrets.
+func GetRepoCloneOptions(ctx context.Context, repoCred *controllerconfig.RepoCredential, kubeClient kubernetes.Client, namespace string, repo *v1alpha1.RepositoryPath) (*git.CloneOptions, error) {
+	if repoCred == nil {
+		return nil, fmt.Errorf("repository credentials cannot be nil")
```
juliev0, line 154: actually, isn't this the expected case if the prefix isn't found? (which is okay)

shubhamdixit863: in that case i shouldn't be returning error in any case just cloneOptions with nil auth ,assuming we have public repositories in CRD ? @juliev0 right?

juliev0: I don't believe we should return an error, right? As stated in the issue: "For any given repository, there may or may not be a matching prefix in there. If there is no matching entry, then no credentials are needed for cloning that RepositoryPath."

config/samples/config.yaml
xdevxy: I think it would be good to put a config map example here for demonstration purpose.

internal/shared/k8/k8_test.go
```diff
@@ -0,0 +1 @@
+package k8
```
xdevxy, line 1: if we don't have any test for this, we can remove this file

internal/shared/git/git.go
```diff
@@ -0,0 +1,252 @@
...
+func CheckGitURL(gitURL string) bool {
+	u, err := url.Parse(gitURL)
+	if err == nil && !Transports.Valid(u.Scheme) {
+		return false
+	}
+	return true
+}
+
+// Parser converts a string into a URL.
+type Parser func(string) (*url.URL, error)
```
xdevxy, line 68: I feel these code can go into a `util` package instead of inside `git`. Can you move it to `util`?

config/samples/config.yaml
xdevxy: Specifically, you can put it under `tests/config/`.

internal/shared/git/git.go
xdevxy: Specifically, you can put it under `internal/shared/util/`.

shubhamdixit863: This basically parses the GitUrls only so I named it to GitUrlParser

xdevxy: I would prefer to put the `GitUrlParser` related logic to a new file like `git_url_parser.go `. But if you like, you can keep if the way it is now. I can do that in the follow up PR.

shubhamdixit863: ok in the followup pr i will fix that

juliev0:
can you resolve the conflicts?

juliev0 (approved):

xdevxy (approved):

PR #120, opened 2024-02-28
chore: Update gitSync API Specs to support source types

### LINKED ISSUE #90: Add Repository type field in GitSync
Add repository type field in GitSync CR which could be like below:

_Originally posted by @xdevxy in https://github.com/numaproj-labs/numaplane/pull/86#discussion_r1488561068_

chandankumar4:
Fixes #90 

### Modifications

Update GitSync API to support source type field and modified as per discussed in above issue.
updated CR will looks like below

```
spec:
  name: numaflow
  path: config/namespace-install
  repoUrl: https://github.com/numaproj/numaflow.git
  targetRevision: main
  kustomize:
    ...
  raw:
    ...
  helm:
    valueFiles:
      - values-1.yaml
      - values-2.yaml
    parameters:
      - name: "service.type"
         value: LoadBalancer
```

### Verification

Tested in local kubernetes cluster


chandankumar4:
Updated Example: https://github.com/numaproj-labs/numaplane-control-manifests/pull/7

api/v1alpha1/gitsync_types.go
```diff
@@ -38,16 +45,42 @@ const (
...
 )
 
 // GitSyncSpec defines the desired state of GitSync
 type GitSyncSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// RepositoryPath lists the Git Repository path to watch
-	RepositoryPath RepositoryPath `json:"repositoryPath"`
+	// Name is a unique name
+	Name string `json:"name"`
```
xdevxy, line 59: We don't need `Name` anymore as we move these field directly under spec from `RepositoryPath`.

api/v1alpha1/gitsync_types.go
```diff
@@ -154,6 +190,32 @@ func (gitSyncSpec *GitSyncSpec) GetDestinationNamespace(cluster string) string {
 	return ""
 }
 
+// ExplicitType returns the type (e.g., Helm, Kustomize, etc.) of the application. If either none or multiple types are defined, returns an error.
+func (gitSyncSpec *GitSyncSpec) ExplicitType() (SourceType, error) {
```
xdevxy, line 194: I think it is better to default to raw if user doesn't specify anything.

api/v1alpha1/gitsync_types.go
chandankumar4: I also thought to remove it but it was there in the [defined schema](https://github.com/numaproj-labs/numaplane/issues/90#issuecomment-1967347677) so I had left it, I'll remove it now

juliev0 (commented):
Did you confirm it works if the explicit type is changed later on? Also, if we go from explicit to implicit type, or implicit to explicit type?

chandankumar4:
> Did you confirm it works if the explicit type is changed later on? Also, if we go from explicit to implicit type, or implicit to explicit type?

I'm trying to understand, can you mention some example like how the GitSync CR will be after change in explicit type ?

juliev0:
> > Did you confirm it works if the explicit type is changed later on? Also, if we go from explicit to implicit type, or implicit to explicit type?
> 
> I'm trying to understand, can you mention some example like how the GitSync CR will be after change in explicit type ?

Sorry, I just meant if the GitSync CR changes from:

- `kustomize` to `raw` (change in explicit type)
- `kustomize` to none specified (which as I understand is going to be considered `raw`)
- none specified to `kustomize`

I didn't look closely at the code. If you're sure it will work but haven't tested it, I'm okay with that.

juliev0:
> > > Did you confirm it works if the explicit type is changed later on? Also, if we go from explicit to implicit type, or implicit to explicit type?
> > 
> > 
> > I'm trying to understand, can you mention some example like how the GitSync CR will be after change in explicit type ?
> 
> Sorry, I just meant if the GitSync CR changes from:
> 
> * `kustomize` to `raw` (change in explicit type)
> * `kustomize` to none specified (which as I understand is going to be considered `raw`)
> * none specified to `kustomize`
> 
> I didn't look closely at the code. If you're sure it will work but haven't tested it, I'm okay with that.

Actually, never mind - it definitely won't work and that's because we never implemented the `Update()` method of GitSyncProcessor to process a modified GitSync spec. And that's because of the refactor that that ended up being on hold. We'll need to make sure that as part of the refactor we account for changes in the GitSync spec.

chandankumar4:
> process a modified GitSync spec.

Right, Update and Delete both needs to be implemented

juliev0 (approved):

xdevxy (approved):

Issue #113, opened 2024-02-22
feat: Managed Cluster Cache


xdevxy:
# Summary

Build a managed cluster cache on top of the ClusterCache provided by gitops engine.

# Use Cases

When would you use this?

A live cluster cache when query always ensured its up to date before returning the state.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

xdevxy:
Closed with #124

Issue #112, opened 2024-02-22
feat: resources tracking


xdevxy:
# Summary

We may set the gitsync annotation with the name of the managing GitSync on all resources that are managed. 

# Use Cases
To be able to identify resources Numaplane manages.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

xdevxy:
Closed with #124

PR #124, opened 2024-03-03
feat: switch to work queue model for syncing

### LINKED ISSUE #46: Worker goroutines to reduce the number of concurrent goroutines reading from RepositoryPaths

xdevxy:
Fixes #46 #112 #113 

### Modifications
This PR switches to work queue model for automated syncing to git state. It also added resource tracking with annotation, managed cluster cache for retrieving the up-to-date resources on the cluster. Mostly importantly, it utilized gitops engine reconciliation and syncing logic underneath. All improvements are marked as TODO to be address in follow up PRs.

### Verification
Verified in a local cluster,
```
make start
kubectl get pipeline -n numaflow-pipeline
```


internal/shared/git/git_test.go
```diff
@@ -180,147 +174,147 @@ func TestGetURLScheme(t *testing.T) {
...
-		},
-	}
-	secret := &v1.Secret{}
-	client.EXPECT().GetSecret(context.Background(), gomock.Any(), gomock.Any()).Return(secret, nil).AnyTimes()
-
-	options, err := GetRepoCloneOptions(context.Background(), cred, client, "testnamespace", "https://github.com/numaproj-labs/numaplane.git")
-	assert.Error(t, err)
-	assert.Nil(t, options)
-}
+//// Testing the case when GitSync CRD has repository path but the RepoCredential Doesn't have the entry for it
```
xdevxy, line 177: Will re-enable the test in a follow up along other tests for git authz.

internal/git/util.go
```diff
@@ -0,0 +1,201 @@
...
+	gitCredentials := gitShared.FindCredByUrl(gitSync.Spec.RepoUrl, globalConfig)
+	cloneOptions, err := gitShared.GetRepoCloneOptions(ctx, gitCredentials, client, namespace, gitSync.Spec.RepoUrl)
+	if err != nil {
+		return nil, fmt.Errorf("error getting  the  clone options: %v", err)
+	}
+	return cloneRepo(ctx, gitSync, cloneOptions)
+}
+
+// GetLatestManifests monitors a Git repository for changes. It fetches updates from the remote repository,
+// checks the latest commit hash against the stored hash in the GitSync object,
```
juliev0, line 38: Is this one a "todo" I guess?

xdevxy: yes, the comment needs to be updated. Thanks!

juliev0 (commented):
For anything that we were previously storing in memory, within our goroutine on the stack, will we have a place to store that (some sort of GitSync object perhaps)?

xdevxy:
> For anything that we were previously storing in memory, within our goroutine on the stack, will we have a place to store that (some sort of GitSync object perhaps)?

you mean the git repository reference? (other things like the commit status will go into GitSync CR.) We can probably consider that as an optimization we can do later, however with current work queue model, each task is considered isolated, and with GitSync CR can be updated, I'm not sure how much value it would be to cache that.

internal/sync/syncer.go
```diff
@@ -0,0 +1,311 @@
...
+		return fmt.Errorf("failed to get the manifest of key %q, %w", key, err)
+	}
+	uns, err := toUnstructuredAndApplyAnnotation(manifests, gitSyncName)
+	if err != nil {
+		return fmt.Errorf("failed to parse the manifest of key %q, %w", key, err)
+	}
+	synced := s.sync(gitSync, uns)
+	if synced {
+		log.Info("GitSync object is successfully synced.")
+	}
```
juliev0, line 223: Does the gitops-engine assume that the full set of manifests passed here represent the complete set of Resources associated with the GitSync? Is that how it know that a manifest has been deleted?

internal/sync/syncer.go
```diff
@@ -0,0 +1,311 @@
...
+		gitSync.Spec.Destination.Namespace,
+		openAPISchema,
+		opts...,
+	)
+	defer cleanup()
+	if err != nil {
+		return false
+	}
+
+	syncCtx.Sync()
```
juliev0, line 276: Do we know if this is doing a Patch or Update under the hood and if there's a ResourceVersion check in the case that the reconciliationResult is now outdated?

internal/git/util.go
juliev0: I believe the comment below about a "ticker" is an old comment as well, right?

internal/shared/common.go
```diff
@@ -0,0 +1,12 @@
...
+
+const (
+	// AnnotationKeyGitSyncInstance Resource metadata annotations (keys and values) used for tracking
+	AnnotationKeyGitSyncInstance = "numaplane.numaproj.io/tracking-id"
+	// SSAManager is the default numaplane manager name used by server-side apply syncs
+	SSAManager = "numaplane-controller"
+	// EnvLogFormat log format that is defined by `--logformat` option
+	EnvLogFormat = "ARGOCD_LOG_FORMAT"
+	// EnvLogLevel log level that is defined by `--loglevel` option
+	EnvLogLevel = "ARGOCD_LOG_LEVEL"
```
juliev0, line 11: remove any instances of "ArgoCD"

internal/sync/syncer.go
```diff
@@ -0,0 +1,311 @@
...
+	// It makes sure each element in the list will be assigned every N milliseconds.
+	for {
+		select {
+		case <-ctx.Done():
+			log.Info("Shutting down synchronizer job assigner.")
+			return nil
+		default:
+			assign()
+		}
+		// Make sure each of the key will be assigned at most every N milliseconds.
```
juliev0, line 146: This is neat logic. I see here you wrote "the key will be assigned at most every N milliseconds" - that makes sense since the Workers could get busy and the channel could get blocked. I wonder if the comment above should also say "at most"?

internal/sync/syncer.go
```diff
@@ -0,0 +1,311 @@
...
+			result := s.options.taskInterval / l
+			if result > 0 {
+				return result
+			}
+			return 1
+		}()))
+	}
+}
+
+// Function run() defines each of the worker's job.
```
juliev0, line 161:
```suggestion
// Function run() defines each worker's job.
```

internal/sync/cache.go
```diff
@@ -0,0 +1,371 @@
...
+
+type ResourceInfo struct {
+	GitSyncName string
+
+	Health       *health.HealthStatus
+	manifestHash string
+}
+
+// LiveStateCache is a cluster caching that stores resource references and ownership
+// references. It also stored custom metadata for resources managed by GitSyncs. It always
```
juliev0, line 73:
```suggestion
// references. It also stores custom metadata for resources managed by GitSyncs. It always
```

internal/sync/cache.go
```diff
@@ -0,0 +1,371 @@
...
+) LiveStateCache {
+	return &liveStateCache{
+		clusterCacheConfig: clusterCacheConfig,
+	}
+}
+
+func (c *liveStateCache) getCluster() clustercache.ClusterCache {
+	c.lock.RLock()
+	if c.cluster != nil {
+		return c.cluster
```
juliev0, line 125: I guess we need to do a `c.lock.RUnlock()` in the case of premature return?

internal/sync/cache.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	cacheSettings := c.cacheSettings
+	c.lock.RUnlock()
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.cluster != nil {
+		return c.cluster
+	}
```
juliev0, line 135: I guess the answer to my previous PR about "why do we re-check for this here?" is because it could've been nil when we checked above, but then set later prior to our entering our single-writer block, huh?

juliev0:
> > For anything that we were previously storing in memory, within our goroutine on the stack, will we have a place to store that (some sort of GitSync object perhaps)?
> 
> you mean the git repository reference? (other things like the commit status will go into GitSync CR.) We can probably consider that as an optimization we can do later, however with current work queue model, each task is considered isolated, and with GitSync CR can be updated, I'm not sure how much value it would be to cache that.

Yeah, I guess we can see if there's anything that comes up to store in memory. One thing we were storing before was `lastCommitHash` so we didn't have to go to the GitSync's CommitStatus to get it. As far as I understand, GitSync Status should be part of the Informer cache anyway, so I suppose that already exists in memory.

internal/git/util.go
```diff
@@ -0,0 +1,201 @@
...
+	} else {
+		// Retrieving the commit object matching the hash.
+		tree, err := getCommitTreeAtPath(r, gitSync.Spec.Path, *hash)
+		if err != nil {
+			return manifests, err
+		}
+		// Read all the files under the path and apply each one respectively.
+		err = tree.Files().ForEach(func(f *object.File) error {
+			logger.Debugw("read file", "file_name", f.Name)
+			// TODO: this currently assumes that one file contains just one manifest - modify for multiple
```
juliev0, line 86: so, this is still not implemented, right? should there be an issue for it?

internal/sync/cache.go
```diff
@@ -0,0 +1,371 @@
...
+	return clusterCache
+}
+
+func (c *liveStateCache) PopulateResourceInfo(un *unstructured.Unstructured, isRoot bool) (interface{}, bool) {
+	res := &ResourceInfo{}
+	c.lock.RLock()
+	settings := c.cacheSettings
+	c.lock.RUnlock()
+
+	res.Health, _ = health.GetResourceHealth(un, settings.clusterSettings.ResourceHealthOverride)
```
juliev0, line 180: do you know how Resource Health is being used in this cache?

internal/sync/syncer.go
juliev0: Now that I say that...say the channel has been blocked for awhile...then we will sleep here even though the task may be due, right? (thinking if we'd ever want to do anything about that in the future or not. I suppose this is an undesirable state of too much work for the Controller with not enough workers, which means somebody should probably tune their Controller for more workers and more CPU. And maybe our sleeping here isn't such a bad thing)

internal/shared/common.go
juliev0: By the way, we currently have a "NUMAPLANE_DEBUG" environment variable which gets set to true or false (if false, then "Info" level assumed). Should we have one environment variable for everything? I haven't looked closely at this logrus logging. So, are we going to have logs of two different formats in our log files or no?

internal/sync/cache.go
xdevxy: It is used to check if we need to skip a resource update when a resource change event is triggered. https://github.com/numaproj-labs/numaplane/pull/124/files#diff-f68311d50ffbcd9fd1106622bfe4fca80b3245c1b4403d2841431600cc9e7129R161

xdevxy:
> > > For anything that we were previously storing in memory, within our goroutine on the stack, will we have a place to store that (some sort of GitSync object perhaps)?
> > 
> > 
> > you mean the git repository reference? (other things like the commit status will go into GitSync CR.) We can probably consider that as an optimization we can do later, however with current work queue model, each task is considered isolated, and with GitSync CR can be updated, I'm not sure how much value it would be to cache that.
> 
> Yeah, I guess we can see if there's anything that comes up to store in memory. One thing we were storing before was `lastCommitHash` so we didn't have to go to the GitSync's CommitStatus to get it. As far as I understand, GitSync Status should be part of the Informer cache anyway, so I suppose that already exists in memory.

Yeah, if it is already in memory we probably do not want to cache it again.

internal/shared/common.go
xdevxy: Yeah, unfortunately gitops engine requires logrus. We can probably have an issue to move to logrus for our normal logging as well.

internal/sync/syncer.go
xdevxy: That is configured by the `replace` sync option.

chandankumar4 (approved):
overall, it looks good
Thanks for refactoring!!!

internal/sync/syncer.go
juliev0, line 146:
```suggestion
		// Make sure each of the keys will be assigned at most every N milliseconds.
```

internal/sync/syncer.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+		e := s.gitSyncList.Front()
+		if key, ok := e.Value.(string); ok {
+			s.gitSyncList.MoveToBack(e)
+			keyCh <- key
+		}
+	}
+
+	// Following for loop keeps calling assign() function to assign watching tasks to the workers.
+	// It makes sure each element in the list will be assigned every N milliseconds.
```
juliev0, line 137:
```suggestion
	// It makes sure each element in the list will be assigned at most every N milliseconds.
```

internal/shared/common.go
juliev0: Sure. Or at least have a single environment variable for both? I'm fine with opening a separate issue for that.

internal/shared/kubernetes/kubernetes.go
```diff
@@ -15,3 +20,59 @@ func IsValidKubernetesNamespace(name string) bool {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get annotations from target object %s %s/%s: %w", un.GroupVersionKind().String(), un.GetNamespace(), un.GetName(), err)
+	}
+	if annotations != nil {
+		return annotations[key], nil
+	}
+	return "", nil
+}
+
+// SetGitSyncInstanceAnnotation the recommended app.kubernetes.io/instance annotation against an unstructured object
```
juliev0, line 36:
```suggestion
// SetGitSyncInstanceAnnotation sets the recommended app.kubernetes.io/instance annotation against an unstructured object
```

juliev0: Regarding the comment on "legacy labeling if environment variable is set" below - want to make sure that it is the case.?

internal/sync/syncer.go
juliev0: So, "replace==false" I guess means it's doing a Patch, right? And that Patch probably doesn't have a ResourceVersion in it, I imagine? So, there's some small risk that between calculating the reconciliationResult and applying it, something changes (say user adds a new field but our reconciliationResult doesn't override it). 

Back when we were doing client-side apply, auto-healing wouldn't take care of this issue because our "lastAppliedConfiguration" would forever be wrong. But since you're using Server Side Apply, auto-healing won't have that problem. We should be able to come back here, take the delta and then see that we need to override the field. So, as far as I can tell we should be good there.

internal/shared/common.go
xdevxy: It seems they are two different use cases. So will leave it there, and create an issue for it.

xdevxy: https://github.com/numaproj-labs/numaplane/issues/127

internal/sync/cache.go
xdevxy: yes.

internal/shared/kubernetes/kubernetes.go
xdevxy: No, removed.

internal/git/util.go
xdevxy: Sure, let me create an issue to track verify if it is still applicable then fix if so.

xdevxy: https://github.com/numaproj-labs/numaplane/issues/128

internal/sync/syncer.go
xdevxy: yeah I think that is an indication of more workers are needed. but lets keep an eye on this to see if improvements needed especially with auto healing.

internal/git/util.go
juliev0: thanks!

juliev0 (approved):

PR #140, opened 2024-03-08
Fix the kustomize and helm dependency and remote changes sync

### LINKED ISSUE #100: Detect kustomization and Helm changes in case it's reference by another repo
# Summary

What change needs making?

# Use Cases

When would you use this?

---

chandankumar4:
Fixes #100 

### Modifications

- Generate manifest from remote repo if referenced inside configuration of Kustomize/Helm
- Sync the changes if there is any change in remote reference, because manifest is generated each time based on latest changes from reference

### Verification

Tested it on local cluster with below below GitSync CR

For kustomize
```
apiVersion: numaplane.numaproj.io/v1alpha1
kind: GitSync
metadata:
  name: gitsync-example
  namespace: numaplane-system
spec:
  path: kustmize-remote-example
  repoUrl: https://github.com/chandankumar4/numaplane-sample.git
  targetRevision: main
  kustomize: {}
  destination:
    cluster: staging-usw2-k8s
    namespace: numaflow-pipeline
``` 

For Helm
```
apiVersion: numaplane.numaproj.io/v1alpha1
kind: GitSync
metadata:
  name: gitsync-example
  namespace: numaplane-system
spec:
  path: helm-dependency
  repoUrl: https://github.com/argoproj/argocd-example-apps.git
  targetRevision: master
  helm: {}
  destination:
    cluster: staging-usw2-k8s
    namespace: numaflow-pipeline
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,13 +38,13 @@ RUN curl --retry 3 --silent --location --remote-name https://get.helm.sh/helm-${
     mkdir -p /tmp/helm && tar -C /tmp/helm -xf helm-${HELM_VERSION}-linux-amd64.tar.gz && \
     install -m 0755 /tmp/helm/linux-amd64/helm /usr/local/bin/helm
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+# Use alpine as minimal base image to package the manager binary
+FROM alpine
 WORKDIR /
 COPY --from=builder /workspace/manager .
 COPY --from=builder /usr/local/bin/kustomize /usr/local/bin/kustomize
 COPY --from=builder /usr/local/bin/helm /usr/local/bin/helm
-USER 65532:65532
+
+RUN apk add --no-cache git
 
 ENTRYPOINT ["/manager"]
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -56,8 +56,6 @@ spec:
       #             operator: In
       #             values:
       #               - linux
-      securityContext:
-        runAsNonRoot: true
         # TODO(user): For common cases that do not require escalating privileges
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -38,13 +38,13 @@ RUN curl --retry 3 --silent --location --remote-name https://get.helm.sh/helm-${
     mkdir -p /tmp/helm && tar -C /tmp/helm -xf helm-${HELM_VERSION}-linux-amd64.tar.gz && \
     install -m 0755 /tmp/helm/linux-amd64/helm /usr/local/bin/helm
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+# Use alpine as minimal base image to package the manager binary
+FROM alpine
```
chandankumar4, line 42: Had to replace this to alpine as git installation is required to build dependencies for kustomize as copying just git binary from builder container is not working.

config/manager/manager.yaml
```diff
@@ -56,8 +56,6 @@ spec:
       #             operator: In
       #             values:
       #               - linux
-      securityContext:
-        runAsNonRoot: true
```
chandankumar4, line 60: need to remove this as alpine container is not a nonroot container

internal/shared/exec/exec.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	defaultTimeout = 10 * time.Second
+	defaultTimeout = 90 * time.Second
```
chandankumar4, line 11: Helm dependencied fetch takes time to had to increase it, also looked in argo-cd that is also using 90s [here](https://github.com/argoproj/argo-cd/blob/3fee8cbf815fb540aec4b0e0a248d28cc84a86a6/util/exec/exec.go#L36)

xdevxy (approved):

Issue #58, opened 2024-01-25
Filter yaml files after reading the manifest repository


chandankumar4:
While reading the manifest from repository, filter the yaml file based on below criteria.
- Filter the files with extension `.yaml` from set of files.
- Filter `Kustomization.yaml` file (If `Kustomization` is fixed name then it can be filtered via file name as well?)

juliev0:
I am kind of thinking maybe we don't filter on name `kustomization.yaml`, but instead we will have an identifier in `RepositoryPath` struct which will identify whether it's a kustomization-type or not.

For this code, let's assume that we only got here if the `RepositoryPath` is set to "raw" type. We can consider that the yaml files could be any type of yaml file, whether it's Kubernetes resource or something else. I'd say once we get the first resource in the file prior to a "---" delimiter, let's try Chandan's apply method with it. If it fails, we assume the entire file does not contain Kubernetes resources, and since this is not explicitly an error maybe we issue a Warn message (or alternatively an Info message). 

How does that sound to you @chandankumar4 ?

juliev0:
> I'd say once we get the first resource in the file prior to a "---" delimiter, let's try Chandan's `ApplyResource()` method with it. If it fails, we assume the entire file does not contain Kubernetes resources...

Actually, we do need to make sure that it failed for the reason that it wasn't a Kubernetes resource...

xdevxy:
If we want to filter based on the extension, we should make sure to filter `.yaml`, `.yml`, `.json` files. As the last two should also be supported. https://kubernetes.io/docs/concepts/overview/working-with-objects/#describing-a-kubernetes-object

juliev0:
> If we want to filter based on the extension, we should make sure to filter `.yaml`, `.yml`, `.json` files. As the last two should also be supported. https://kubernetes.io/docs/concepts/overview/working-with-objects/#describing-a-kubernetes-object

good point

PR #141, opened 2024-03-11
filtering yaml and json files

shubhamdixit863:
Fixes https://github.com/numaproj-labs/numaplane/issues/58

### Modifications

[IsValidKubernetesManifestFile](https://github.com/numaproj-labs/numaplane/blob/feat/filterYamlfiles/internal/shared/kubernetes/kubernetes.go#L81) function has been added which checks for the valid yaml and json files in the syncing repo from Gitsync CRD

### Verification

Unit tests have been written for the above [method](https://github.com/numaproj-labs/numaplane/blob/feat/filterYamlfiles/internal/shared/kubernetes/kubernetes_test.go#L68) 

This function tests the `IsValidKubernetesManifestFile` method to ensure it correctly identifies valid Kubernetes manifest file names. It checks multiple file names with different extensions (`.md`, `.yml`, `.json`, `.yaml`, `.go`) and verifies whether the method accurately classifies each as either a valid Kubernetes manifest (true) or not (false), based on their file extensions.

### DIFF
--- a/internal/shared/kubernetes/kubernetes.go
+++ b/internal/shared/kubernetes/kubernetes.go
@@ -81,5 +81,5 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
 func IsValidKubernetesManifestFile(fileName string) bool {
 	fileExt := strings.Split(fileName, ".")
 	validExtName := []string{"yaml", "yml", "json"}
-	return slices.Contains(validExtName, fileExt[1])
+	return slices.Contains(validExtName, fileExt[len(fileExt)-1])
 }
--- a/internal/shared/kubernetes/kubernetes.go
+++ b/internal/shared/kubernetes/kubernetes.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"slices"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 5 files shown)


internal/shared/kubernetes/kubernetes.go
```diff
@@ -75,3 +77,9 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
 	}
 	return secret, nil
 }
+
+func IsValidKubernetesManifestFile(fileName string) bool {
+	fileExt := strings.Split(fileName, ".")
+	validExtName := []string{"yaml", "yml", "json"}
+	return slices.Contains(validExtName, fileExt[1])
```
xdevxy, line 84: What if the file name contains multiple `.`?

internal/shared/kubernetes/kubernetes_test.go
```diff
@@ -64,3 +64,48 @@ func TestGetGitSyncInstanceAnnotationWithInvalidData(t *testing.T) {
...
+		},
+
+		{
+			name:         "Valid name yaml",
+			resourceName: "pipeline.yaml",
+			expected:     true,
+		},
+		{
+			name:         "Invalid File",
+			resourceName: "main.go",
```
xdevxy, line 100: can you add a test case to cover the above case I mentioned, e.g. `abc.xyz.yaml`

juliev0:
@xdevxy once you are good with this one, I don't need to review

xdevxy (approved):

PR #142, opened 2024-03-11
Add Make manifests command to generate install.yaml

### LINKED ISSUE #121: Makefile Target to auto-generate install.yaml
# Summary

We need a target similar to Numaflow's "make manifests" to generate an [install.yaml file](https://github.com/numaproj/numaflow/blob/v1.1.6/Makefile#L210) to persist in our repository. (once we have Github version tags, we can grab the correct file pertaining to the version)

One difference from Numaflow is that our file will not be a complete deployable file on its own. This is because we require the [cluster name to be included in our ConfigMap](https://github.com/numaproj-labs/numaplane/blob/b2468d51bcab82dfef8d918a15f5d031cf40c708/config/manager/config.yaml#L8). The idea is that the organization using this manifest is expected to set this themselves as part of their kustomization built on top of this file.

We do still need to be able to do tests locally as part of "make start", so we will need to set the field in the ConfigMap as part of that (as we do [here](https://github.com/numaproj-labs/numaplane/blob/95066f30e304d343a592fa83ba71ab2279669d1b/Makefile#L157)). We should remove that field from our main ConfigMap spec, and instead separate it into a ConfigMap Patch file which only gets run when testing locally.

This Makefile target should be called from (or a dependency of) "make codegen" - which is an Issue defined [here](https://github.com/numaproj-labs/numaplane/issues/78).

# Use Cases

When would you use this?

---

chandankumar4:
Fixes #121 

### Modifications

Add "make manifests" command to generate `install.yaml`

### Verification

Verified by applying install.yaml in local k8s cluster

### DIFF
--- a/config/manager/config.yaml
+++ b/config/manager/config.yaml
@@ -6,7 +6,6 @@ metadata:
   namespace: numaplane-system
 data:
   config.yaml: |
-    clusterName: "CLUSTER_NAME_VALUE" # Replace with the cluster name where numaplane will get deployed.
     timeIntervalSec: 60
     repoCredentials:
     - url: "github.com/numaproj-labs"
--- a/config/manager/config.yaml
+++ b/config/manager/config.yaml
@@ -7,26 +7,3 @@ metadata:
 data:
   config.yaml: |
     timeIntervalSec: 60
-    repoCredentials:
-    - url: "github.com/numaproj-labs"
-      httpCredential:
-        username: "exampleUser"
-        password:
-          name: "http-creds"
-          key: "password"
-    - url: "github.com:numaproj"
-      sshCredential:
-        SSHKey:
-          name: "ssh-creds"
-          key: "sshKey"
-      tls:
-        insecureSkipVerify: true
-    - url: "github.com/numalabs"
-      httpCredential:
-        username: "exampleuser3"
-        password:
-          name: "http-creds"
-          key: "password"
-      tls:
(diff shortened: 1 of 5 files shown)


config/manager/config.yaml
```diff
@@ -6,7 +6,6 @@ metadata:
   namespace: numaplane-system
 data:
   config.yaml: |
-    clusterName: "CLUSTER_NAME_VALUE" # Replace with the cluster name where numaplane will get deployed.
```
chandankumar4, line 9: we are using a custom config for `make start`, so we don't need to patch anything for local

config/manager/config.yaml
```diff
@@ -6,7 +6,6 @@ metadata:
...
     timeIntervalSec: 60
     repoCredentials:
     - url: "github.com/numaproj-labs"
```
chandankumar4, line 11: Should we remove this complete data from config map? as this is for testing purpose and not for prod and user should add these value before deploying

juliev0: good point, yes, the repoCredentials shouldn't be in there either. I kind of think we can leave the `timeIntervalSec` in there and just consider that to be the default.

chandankumar4: okay, Made the changes accordingly

juliev0 (approved):

xdevxy (approved):

Issue #65, opened 2024-01-29
OwnerReference on applied resources


juliev0:
We should have an OwnerReference (to the GitSync) on any resources we apply, so that if the GitSync were deleted, then all of the resources would automatically be deleted as well.

xdevxy:
As cross-namespace owner references are disallowed by design, we will not proceed this. Closed this issue for now.

PR #143, opened 2024-03-11
Ownership reference to the kubernetes manifests

shubhamdixit863:
Fixes  https://github.com/numaproj-labs/numaplane/issues/65

### Modifications

[ApplyOwnershipReference](https://github.com/numaproj-labs/numaplane/blob/feat/ownerReference/internal/shared/kubernetes/kubernetes.go#L88) method has been added and called in the [syncer.go](https://github.com/numaproj-labs/numaplane/blob/feat/ownerReference/internal/sync/syncer.go#L219) to apply the ownership reference to kubernetes manisfests

### Verification

Unit tests Have been added to check for both yaml and json mainfests
Three cases have been covered adding the ownership reference to
1- yaml
2- json
3- Updating the existing ownership reference

For Yaml
 [ApplyOwnershipReference](https://github.com/numaproj-
labs/numaplane/blob/feat/ownerReference/internal/shared/kubernetes/kubernetes_test.go#L71) has been added to check if ownership reference gets applied for yaml
And 
For [Json](https://github.com/numaproj-labs/numaplane/blob/feat/ownerReference/internal/shared/kubernetes/kubernetes_test.go#L211) 

These unit tests checks  if the method correctly adds  an ownerReferences field to a Kubernetes Deployment resource, indicating ownership by a GitSync custom resource. The test verifies the modified resource string and asserts there are no errors.

For  Updating the existing ownership [reference](https://github.com/numaproj-labs/numaplane/blob/feat/ownerReference/internal/shared/kubernetes/kubernetes_test.go#L148) -

This function tests the `ApplyOwnershipReference` method to ensure it correctly appends an ownership reference to an existing list of owner references in a Kubernetes Pod resource. It verifies that the GitSync custom resource is added without affecting existing owner references and checks that the modified resource string is correct, asserting no errors.


internal/shared/kubernetes/kubernetes.go
```diff
@@ -75,3 +83,33 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
 	}
 	return secret, nil
 }
+
+// ApplyOwnershipReference adds  or updates ownerships to kubernetes resources in GitSync CRD
+func ApplyOwnershipReference(manifest string, gitSync *v1alpha1.GitSync) ([]byte, error) {
```
juliev0, line 88: nit: Can we maybe call it `ApplyGitSyncOwnership()` so it's more clear?

internal/shared/kubernetes/kubernetes.go
```diff
@@ -75,3 +83,33 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
 	}
 	return secret, nil
 }
+
+// ApplyOwnershipReference adds  or updates ownerships to kubernetes resources in GitSync CRD
```
juliev0, line 87:
```suggestion
// ApplyOwnershipReference sets the GitSync as the Owner of the Resource through an OwnerReference
```

internal/shared/kubernetes/kubernetes.go
```diff
@@ -75,3 +83,33 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
...
+		APIVersion:         gitSync.APIVersion,
+		Kind:               gitSync.Kind,
+		Name:               gitSync.Name,
+		UID:                gitSync.UID,
+		Controller:         ptr.To(true),
+		BlockOwnerDeletion: ptr.To(true),
+	}
+	existingRef := obj.GetOwnerReferences()
+	existingRef = append(existingRef, ownerRef)
+	obj.SetOwnerReferences(existingRef)
```
juliev0, line 108: I want to make sure we don't possibly add the same OwnerReference when it's already in there. If there's any possibility, then we should check before we add it.

internal/shared/kubernetes/kubernetes.go
xdevxy: This also actually appends to existing owner if already exists, right? may want to make sure the comment also covers that.

`// ApplyOwnershipReference sets the GitSync as the Owner (or one of the Owners if ones already exist) of the Resource through an OwnerReference`

internal/shared/kubernetes/kubernetes.go
xdevxy: Also a test case for that

internal/sync/syncer.go
```diff
@@ -210,10 +210,18 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
 		return fmt.Errorf("failed to clone the repo of key %q, %w", key, err)
 	}
 	manifests, err := git.GetLatestManifests(ctx, repo, gitSync)
+
 	if err != nil {
 		return fmt.Errorf("failed to get the manifest of key %q, %w", key, err)
 	}
-	uns, err := toUnstructuredAndApplyAnnotation(manifests, gitSyncName)
+	// Applying ownership reference
+
```
xdevxy, line 218: nit: the white space should be before the comment.

xdevxy (commented):
Have you verified this in a local cluster?

internal/shared/kubernetes/kubernetes.go
```diff
@@ -75,3 +83,33 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
...
 }
+
+// ApplyOwnershipReference adds  or updates ownerships to kubernetes resources in GitSync CRD
+func ApplyOwnershipReference(manifest string, gitSync *v1alpha1.GitSync) ([]byte, error) {
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	decoded, _, err := decode([]byte(manifest), nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	obj, ok := decoded.(metav1.Object)
```
xdevxy, line 94: Just to unify the code path and avoid errors here, I would recommend to use same pattern how we set [annotation](https://github.com/numaproj-labs/numaplane/blob/main/internal/shared/kubernetes/kubernetes.go#L37-L48), you can use `Unstructured.SetOwnerReferences`.

juliev0: I was looking at that. If we're going to do that, I wonder if we should also avoid unmarshaling yaml twice? What do you think @xdevxy ?

shubhamdixit863:
> Have you verified this in a local cluster?

Yes its has been verified in the local cluster ,ithe ownership has been applied correctly

juliev0:
Thanks @shubhamdixit863 . Ultimately, the best test of this will be to delete the GitSync itself and then verify that the respective resources in fact get deleted.

juliev0:
> > Thanks @shubhamdixit863 . Ultimately, the best test of this will be to delete the GitSync itself and then verify that the respective resources in fact get deleted.
> 
> Oh ,ok that makes sense ,that i will test as well

thanks so much!

internal/shared/kubernetes/kubernetes.go
xdevxy: I saw the new changes didn't take the suggestion, but I'm good with it as I don't feel strong about this.

internal/util/kubernetes/kubernetes_test.go
```diff
@@ -154,3 +368,67 @@ func TestIsValidKubernetesManifestFile(t *testing.T) {
...
+    uid: <uid-of-my-deployment>
+  - apiVersion: v1
+    blockOwnerDeletion: true
+    controller: false
+    kind: ConfigMap
+    name: my-configmap
+    uid: <uid-of-my-configmap>
+  - apiVersion: v1
+    blockOwnerDeletion: true
+    controller: false
```
xdevxy, line 428: I don't quite follow why controller can be false in the case you explicitly set to true at https://github.com/numaproj-labs/numaplane/pull/143/files#diff-3d19777dfd8078f46a7e12b70830cbbb7f5855463b120852fd92c0a30120ffa5R124

xdevxy: Hmm, actually are you testing if the uid already exist then it will not be added again? If so, the test name should be updated to `TestApplyOwnerShipReferenceAlreadyExists`, right?

internal/util/kubernetes/kubernetes_test.go
```diff
@@ -70,7 +73,218 @@ func TestGetGitSyncInstanceAnnotationWithInvalidData(t *testing.T) {
...
+        memory: 128Mi
+      requests:
+        cpu: 250m
+        memory: 64Mi
+`, string(reference))
+	assert.NoError(t, err)
 
+}
+
+func TestApplyOwnerShipReferenceAppendExisting(t *testing.T) {
```
xdevxy, line 151: This name should be `TestApplyOwnerShipReferenceAppendIfNotExisting`?

internal/util/kubernetes/kubernetes_test.go
shubhamdixit863: > I don't quite follow why controller can be false in the case you explicitly set to true at

it was just unit tests ,so i dint much focus on the values ,as I am just testing whether the ownership get appended or not

shubhamdixit863: > I saw the new changes didn't take the suggestion, but I'm good with it as I don't feel strong about this.

I have used  meta v1 unstructured as suggested by you [here]( https://github.com/numaproj-labs/numaplane/blob/54f447bbcabb7285efd8b452253d11ae397aa9f4/internal/util/kubernetes/kubernetes.go#L109) did I miss something ? @xdevxy

Issue #133, opened 2024-03-07
chore: commit the status after each sync


xdevxy:
# Summary

We need to commit the syncing status after each sync.

# Use Cases

Up to date status is shown in GitSync CR after each sync.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #145, opened 2024-03-11
chore: skip sync if states match

### LINKED ISSUE #136: When the git state matches with cluster state, then skip the syncing
# Summary

During git reconciliation, skip syncing if the git desired state matches with the live cluster state.

# Use Cases

Optimization for automated syncing.

---

xdevxy:
Fixes #136 

### Modifications
This PR skip the syncing if the target states in git matches the live states. When comparing the diff, it utilizes the `ThreeWayDiff` calculation from gitops-engine. It also provides ignore options to ignore the fields in the manifest when comparing.

### Verification

- Unit test.
- In local cluster, verified if the manifest doesn't change, then the syncing is skipped.


internal/sync/syncer.go
```diff
@@ -233,14 +234,36 @@ func (r *resourceInfoProviderStub) IsNamespaced(_ schema.GroupKind) (bool, error
 	return false, nil
 }
 
-func (s *Syncer) sync(gitSync *v1alpha1.GitSync, targetObjs []*unstructured.Unstructured) bool {
-	// TODO: if the states match, then skip the syncing.
+func (s *Syncer) sync(gitSync *v1alpha1.GitSync, targetObjs []*unstructured.Unstructured, logger *zap.SugaredLogger) bool {
```
juliev0, line 237: can we add a comment to this function to say what the return boolean indicates? I assume error?

internal/sync/diff.go
```diff
@@ -0,0 +1,226 @@
...
+	//	//diff.WithLogr(),
+	//	diff.WithNormalizer(diffNormalizer),
+	//	//diff.IgnoreAggregatedRoles(diffConfig.IgnoreAggregatedRoles()),
+	//	//diff.WithStructuredMergeDiff(diffConfig.StructuredMergeDiff()),
+	//	//diff.WithGVKParser(diffConfig.GVKParser()),
+	//	//diff.WithManager(diffConfig.Manager()),
+	//	diff.WithServerSideDiff(true),
+	//	diff.WithServerSideDryRunner(diff.NewK8sServerSideDryRunner(resourceOps)),
+	//}
+	array, err := diff.DiffArray(configs, lives, diffOpts...)
```
juliev0, line 221: why is all of this commented out? does it need to be cleaned up? Or is it code we're considering putting back upon more investigation?

xdevxy: sorry that needs to be cleaned up.

internal/sync/syncer.go
xdevxy: Actually do you mind we do it in the PR for https://github.com/numaproj-labs/numaplane/issues/133, the `sync` method will be updated accordingly there.

internal/sync/syncer.go
```diff
@@ -233,14 +234,36 @@ func (r *resourceInfoProviderStub) IsNamespaced(_ schema.GroupKind) (bool, error
...
+	}
+
+	modified, err := StateDiffs(reconciliationResult.Target, reconciliationResult.Live, overrides, diffOpts)
+	if err != nil {
+		return false
+	}
+
+	// If the live state match the target state, then skip the syncing.
+	if !modified.Modified {
+		logger.Info("GitSync object is successfully already synced, skip the syncing.")
```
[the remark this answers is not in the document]
xdevxy: yes

internal/sync/syncer.go
juliev0: so, you're saying the signature of this method will change anyway?

xdevxy: yes

juliev0: okay

juliev0 (approved):

PR #152, opened 2024-03-13
chore: cleanup and add codegen

### LINKED ISSUE #78: "make codegen"
We should have a single Makefile Target - "make codegen" - similar to Numaflow so that it can be called both locally and in the CI. 

This can generate everything that's at least currently being autogenerated based on our GitSync spec, so this just may just be a matter of shuffling around dependencies in the Makefile, in addition to updating our CI to include something like Numaflow's [codegen](https://github.com/numaproj/numaflow/blob/v1.1.6/.github/workflows/ci.yaml#L31) step.

The user should themselves call `make codegen` as part of submitting a PR. And the CI should also call it and verify that the result doesn't differ from the PR (or otherwise it means the user didn't run it).

Note I have broken off [this related item](https://github.com/numaproj-labs/numaplane/issues/121) into its own Issue.

whynowy:
Clean up some manifests and add `codegen` CI.

Fixes #78 

### Modifications

1. Stop using kubebuilder to generate partial rbac config.
2. Update `make docker-build` to `make image` - to be consistent with numaflow.
3. Add `make codegen`, and set up a CI job to make sure the code is up-to-date.
4. Clean up kubebuilder generated manifests, remove some redundant labels.
5. Remove `namespace` spec from installation manifests - different clusters might have different ways/privilege to manage namespaces.


config/default/kustomization.yaml
```diff
@@ -1,139 +1,13 @@
-# Adds namespace to all resources.
-namespace: numaplane-system
-
-# Value of this field is prepended to the
-# names of all resources, e.g. a deployment named
-# "wordpress" becomes "alices-wordpress".
-# Note that it should also match with the prefix (text before '-') of the namespace
-# field above.
-namePrefix: numaplane-
```
whynowy, line 9: To avoid unexpected prefix in case someone does not know.

juliev0:
This is the "make codegen" issue: https://github.com/numaproj-labs/numaplane/issues/78

config/install.yaml
```diff
@@ -234,22 +221,16 @@ kind: ServiceAccount
...
   labels:
     app.kubernetes.io/component: rbac
-    app.kubernetes.io/created-by: numaplane
-    app.kubernetes.io/instance: controller-manager-sa
-    app.kubernetes.io/managed-by: kustomize
     app.kubernetes.io/name: serviceaccount
     app.kubernetes.io/part-of: numaplane
-  name: numaplane-controller-manager
+  name: numaplane-sa
   namespace: numaplane-system
```
juliev0, line 227: Do we need to remove all of the instances of `numaplane-system` namespace in `config` directory?

juliev0 (changes requested):
commented inline

whynowy:
> commented inline

What do you expect to change?

config/install.yaml
juliev0: Can you clarify? I thought you were extracting the namespace from the install.yaml file in the repository so that somebody could use their own namespace instead. For example, when Chandan is testing in the cluster right now, he's using a different namespace.

whynowy: It would be better to exclude the namespace object definition from the manifests, because ppl might have different approach to create namespace. Just like what we do in workflows:

```
kubectl create namespace argo
kubectl apply -n argo -f https://github.com/argoproj/argo-workflows/releases/download/v3.5.5/install.yaml
```

But it's okay to put the namespace definition in the k8s objects (int the manifests), since we prefer them to install in `numaplane-system` namespace. If anyone wants to install in a different namespace, then can simply override it in the kustomzation.yaml file.

juliev0: Okay. If that's how we're doing it with numaflow, I'm okay with making it the same I guess.

juliev0:
@chandankumar4 do you want to review?

go.sum
```diff
@@ -24,6 +24,8 @@ github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be/go.mod h1:ySMOLuW
 github.com/antihax/optional v1.0.0/go.mod h1:uupD/76wgC+ih3iEmQUL+0Ugr19nfwCT1kdvxnR2qWY=
 github.com/argoproj/argo-cd/v2 v2.10.2 h1:rEoW5Aq9EN55674OuPt8YMtpnpen9fvJSVnlA/HUhbo=
 github.com/argoproj/argo-cd/v2 v2.10.2/go.mod h1:nujAuswdQvB6yWI8HubQjfUiLdiIlKlG0ihx2Ht1D28=
```
chandankumar4, line 26: I believe these two line should go away, can you double check it once  running `go mod tidy`

chandankumar4 (approved):
Looks good, just one small comment

go.sum
whynowy: This is good finding!

Fixed and add `go mod tidy` to `codegen` CI.

juliev0 (approved):

PR #157, opened 2024-03-15
Agent initial

juliev0:
Fixes #TODO


internal/controller/config/config.go
```diff
@@ -33,36 +33,10 @@ func GetConfigManagerInstance() *ConfigManager {
...
-}
-
-type TLS struct {
-	InsecureSkipVerify bool `json:"insecureSkipVerify"`
-}
-
-type SecretKeySelector struct {
-	corev1.ObjectReference `mapstructure:",squash"` // for viper to correctly parse the config
-	Key                    string                   `json:"key" `
-	Optional               *bool                    `json:"optional,omitempty" `
```
juliev0, line 65: all of this moved into api/v1alpha1/shared_types.go

internal/controller/config/config.go
```diff
@@ -33,36 +33,10 @@ func GetConfigManagerInstance() *ConfigManager {
 // supposed to be populated from the configmap attached to the
 // controller manager.
 type GlobalConfig struct {
-	ClusterName     string `json:"clusterName"`
-	TimeIntervalSec uint   `json:"timeIntervalSec"`
+	ClusterName     string `json:"clusterName" mapstructure:"clusterName"`
+	TimeIntervalSec uint   `json:"timeIntervalSec" mapstructure:"timeIntervalSec"`
```
juliev0, line 37: apparently, the "json" tags were working, but supposedly viper uses "mapstructure" tags. Adding both won't hurt anything.

api/v1alpha1/gitsync_types.go
```diff
@@ -31,10 +31,6 @@ type GitSyncPhase string
 
 type ConditionType string
 
-// +kubebuilder:validation:Enum=Helm;Kustomize;Raw
-// ApplicationSourceType specifies the type of the application source
-type SourceType string
-
```
juliev0, line 37: all this stuff moved to shared_types.go

api/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	appType := appTypes[0]
+	return appType, nil
+}
+
+type RepoCredential struct {
+	URL            string          `json:"url" mapstructure:"url"`
+	HTTPCredential *HTTPCredential `json:"httpCredential,omitempty" mapstructure:"httpCredential,omitempty"`
+	SSHCredential  *SSHCredential  `json:"sshCredential,omitempty" mapstructure:"sshCredential,omitempty"`
+	TLS            *TLS            `json:"tls,omitempty" mapstructure:"tls,omitempty"`
```
juliev0, line 125: these didn't have omitempty before - which actually seemed to work okay - but I think we are technically supposed to have them for pointer types

internal/controller/config/config.go
```diff
@@ -92,10 +66,12 @@ func (cm *ConfigManager) LoadConfig(onErrorReloading func(error), configPath, co
 	v.OnConfigChange(func(e fsnotify.Event) {
 		cm.lock.Lock()
 		defer cm.lock.Unlock()
-		err = v.Unmarshal(cm.config)
+		newConfig := GlobalConfig{}
+		err = v.Unmarshal(&newConfig)
 		if err != nil {
 			onErrorReloading(err)
 		}
+		cm.config = &newConfig
```
juliev0, line 74: when you call `v.Unmarshal()` it doesn't first clear out the memory of the original object, so if you use the same memory as before, the old fields can still be in there...therefore creating new memory

api/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,159 @@
+/*
+Copyright 2023.
```
xdevxy, line 2: nit: should it be 2024 now? :)

api/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,159 @@
...
+type SingleClusterGenerator struct {
+	Values map[string]string `json:"values" mapstructure:"values"`
+}
+
+// multiple files containing key/value pairs: subsequent entries can override earlier entries
+type MultiClusterFileGenerator struct {
+	Files []*MultiClusterFile `json:"files" mapstructure:"files"`
+}
+
+type MultiClusterFile struct {
```
xdevxy, line 156: nit: add a comment

cmd/main.go
```diff
@@ -29,6 +29,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
```
xdevxy, line 32: nit: remove extra line

internal/controller/config/config.go
xdevxy: Good catch, do we have a test case that will fail without your change?

internal/agent/config.go
```diff
@@ -0,0 +1,129 @@
...
+type KVGenerator struct {
+	Embedded *apiv1.SingleClusterGenerator `mapstructure:"embedded,omitempty"`
+
+	Reference *apiv1.MultiClusterFileGenerator `mapstructure:"reference,omitempty"`
+}
+
+type ConfigManager struct {
+	config *AgentConfig
+	lock   *sync.RWMutex
+	// revisionIndex increments each time the Config changes so the caller can keep track of when
```
xdevxy, line 52: Why do we need to keep track of this ourselves?

api/v1alpha1/shared_types.go
juliev0: I believe this came up in Numaflow as well and the decision was to keep it the same year. I can see some files in numaflow that still say 2022 even though they're newer than that.

internal/agent/config.go
juliev0: we can avoid reprocessing the file each time by only processing it when it changed

api/v1alpha1/shared_types.go
juliev0: I realized the struct was more generic so I renamed it to `CredentialedGitLocation` and added a comment for it there

internal/controller/config/config.go
juliev0: doing that now

juliev0: just added one, and tried the same test using the "main" branch to verify it would fail

juliev0: darn, need to see why it's failing the race test in the CI...

juliev0: fixed the race condition, which was caused by calling `v.OnConfigChange()` after `v.WatchConfig()`

juliev0:
addressed questions and ready for re-review

internal/agent/agent_syncer.go
```diff
@@ -0,0 +1,153 @@
...
+
+// determine if Config was updated and if so, get latest
+// return if new
+func (syncer *AgentSyncer) checkConfigUpdate() bool {
+	configManager := GetConfigManagerInstance()
+	var err error
+	var newRevision int
+	// Reload our copy of the Config if it changed (or load it the first time upon starting)
+	if configManager.GetRevisionIndex() > syncer.configRevision {
+		syncer.config, newRevision, err = configManager.GetConfig()
```
xdevxy, line 98: Why can't we just call GetConfig each time when we need the config? it will be always the latest in that case.

juliev0: For one, it would mean creating a deep copy of the Config, which we don't need to do. But moreover, it would mean having to create a new KVSource based on the generator, which could potentially be time-intensive, such as the case of the `MultiClusterFileKVSource` might be (or if it's not, then another future one could be).

xdevxy:
Approved, as this is the first skeleton for agent. as long as it doesn't affect existing functionality, I'm good with it.

PR #158, opened 2024-03-15
chore: logging fix

whynowy:
Fix logging in main.go

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.20 as builder
+FROM golang:1.21 as builder
 ARG TARGETOS
 ARG TARGETARCH
 ARG KUSTOMIZE_VERSION="v5.3.0"
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -20,16 +20,17 @@ import (
 	"context"
 	"flag"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
-	// to ensure that exec-entrypoint and run can make use of them.
+	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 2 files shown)


cmd/main.go
```diff
@@ -110,34 +108,43 @@ func main() {
...
+	if err != nil {
+		logger.Fatalw("Unable to set GitSync controller", zap.Error(err))
+	}
 
+	// Watch GitSync objects
+	if err := gitSyncController.Watch(&source.Kind{Type: &apiv1.GitSync{}}, &handler.EnqueueRequestForObject{},
+		predicate.Or(
+			predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{},
+		)); err != nil {
+		logger.Fatalw("Unable to watch GitSyncs", zap.Error(err))
```
juliev0, line 135: Interesting. I'd actually added the GenerationChanged Predicate in [this](https://github.com/numaproj-labs/numaplane/pull/54/files) PR but I guess it was removed at some point.

juliev0: Do you have any type of Label specifically in mind as far as watching for Label changes as well?

juliev0: Actually, regarding my first comment, I noticed the code is still in there: `reconciler.SetupWithManager()` does actually look for the `GenerationChangedPredicate{}` still

whynowy: Reverted!

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.20 as builder
+FROM golang:1.21 as builder
```
juliev0, line 2: I guess that would've also taken care of [this](https://github.com/numaproj-labs/numaplane/pull/149) issue, right? So, now we should be good to use go 1.21 as long as we all are running that version locally

juliev0 (approved):

Issue #101, opened 2024-02-16
Put any manifests we rely on for testing in a controlled location in our repo


juliev0:
# Summary

gitsync_processor_test.go relies on the file structure in numaflow repository as well as the file structure in this repo not changing. It would be nice to have a location in our repository where we can put any files we need for these unit tests, which is guaranteed not to change. 

@xdevxy and I agreed that any test files for unit tests can go in the tests/manifests directory. This can include both GitSync spec files that we need to pull from Github as well as any the manifest files referenced by the RepositoryPaths.

(Note, the exception is if we have any tests that require authorization, then we won't be able to do this for that test.)

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
@shubhamdixit863 I just updated the description - thanks

PR #159, opened 2024-03-18
moved the test config to manifests folder

shubhamdixit863:
Yes

Fixes https://github.com/numaproj-labs/numaplane/issues/101

### Modifications
Modified the config.yaml file  in tests/manifests folder with the latest fields
Modified config tests and the controller tests to use the config file from tests/manifests folder

### Verification

As no new feature has been added ,only the existing tests cases have been checked if they run properly


juliev0:
Sorry, I think this Issue maybe wasn't clear. I was specifically hoping our tests could no longer reference repos like:
- https://github.com/numaproj-labs/numaplane-control-manifests
- https://github.com/numaproj/numaflow
And we could just put all of the files that we need for these tests within this repo instead, in `tests/manifests`.

shubhamdixit863:
> Sorry, I think this Issue maybe wasn't clear. I was specifically hoping our tests could no longer reference repos like:
> 
> - https://github.com/numaproj-labs/numaplane-control-manifests
> 
> - https://github.com/numaproj/numaflow
> 
> And we could just put all of the files that we need for these tests within this repo instead, in `tests/manifests`.

ok i will modify that .

shubhamdixit863:
@juliev0 i have checked we are not using any resources  as of now from these git urls ,we are just using these urls for performing gitsync  tests in util_test.go . So I believe we dont have any dependency on these external repos

juliev0:
> @juliev0 i have checked we are not using any resources as of now from these git urls ,we are just using these urls for performing gitsync tests in util_test.go . So I believe we dont have any dependency on these external repos

This is an example of a dependency on the numaflow repo: https://github.com/numaproj-labs/numaplane/blob/main/internal/git/util_test.go#L279 Let's say that `config/namespace-install` directory changes...

shubhamdixit863:
No longer needed

PR #160, opened 2024-03-18
Add commit status after syncing

### LINKED ISSUE #133: chore: commit the status after each sync

xdevxy:
Fixes #133 

### Modifications

This PR adds commit status after each sync even when there is a sync failure.

### Verification
Verified in a local cluster.

### DIFF
--- a/internal/sync/syncer.go
+++ b/internal/sync/syncer.go
@@ -10,12 +10,15 @@ import (
 
 	"github.com/argoproj/gitops-engine/pkg/diff"
 	gitopsSync "github.com/argoproj/gitops-engine/pkg/sync"
+	gitopsSyncCommon "github.com/argoproj/gitops-engine/pkg/sync/common"
 	kubeUtil "github.com/argoproj/gitops-engine/pkg/utils/kube"
 	log "github.com/sirupsen/logrus"
 	"go.uber.org/zap"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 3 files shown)


internal/sync/syncer.go
```diff
@@ -234,18 +243,25 @@ func (r *resourceInfoProviderStub) IsNamespaced(_ schema.GroupKind) (bool, error
 	return false, nil
 }
 
-func (s *Syncer) sync(gitSync *v1alpha1.GitSync, targetObjs []*unstructured.Unstructured, logger *zap.SugaredLogger) bool {
+// sync compares the live state of the watched resources to target state defined in git
+// for the given GitSync. If not matches, syncing the state to the live objects. Otherwise,
```
juliev0, line 247:
```suggestion
// for the given GitSync. If it doesn't match, sync the state to the live objects. Otherwise,
```

internal/sync/syncer.go
```diff
@@ -335,3 +351,41 @@ func toUnstructuredAndApplyAnnotation(manifests []string, gitSyncName string) ([
 	}
 	return uns, nil
 }
+
+// updateCommitStatus will update the commit status in git sync CR also if
+// error happened while syncing the target state then it update the error
+// reason as well.
```
juliev0, line 357:
```suggestion
// updateCommitStatus will update the commit status in git sync CR.
// If an error occurred while syncing the target state, then it updates the error
// reason as well.
```

internal/sync/syncer.go
```diff
@@ -335,3 +351,41 @@ func toUnstructuredAndApplyAnnotation(manifests []string, gitSyncName string) ([
...
+		if apierrors.IsNotFound(err) {
+			return nil
+		} else {
+			logger.Errorw("Unable to get GitSync", "err", err)
+			return err
+		}
+	}
+
+	gitSync.Status.CommitStatus = &commitStatus
+	if err := kubeClient.Status().Update(ctx, gitSync); err != nil {
```
juliev0, line 386: so, I guess in the case of failing to update either because of network issue with K8S API, or resourceVersion conflict, we probably won't retry ever, right?

juliev0: If we did do exponential backoff here, it would only tie up one worker goroutine, right? and should be rare? What are your thoughts?

xdevxy: I think it is fine to not retry with exponential backoff here as it will be picked up by another worker in next round. I don't see in numaflow we are doing retry if status update failed.

internal/sync/syncer.go
```diff
@@ -234,18 +243,25 @@ func (r *resourceInfoProviderStub) IsNamespaced(_ schema.GroupKind) (bool, error
...
+	gitSync *v1alpha1.GitSync,
+	targetObjs []*unstructured.Unstructured,
+	logger *zap.SugaredLogger,
+) (gitopsSyncCommon.OperationPhase, string) {
 	logEntry := log.WithFields(log.Fields{"gitsync": gitSync})
 
 	reconciliationResult, modified, err := s.compareState(gitSync, targetObjs)
 	if err != nil {
-		return false
+		return gitopsSyncCommon.OperationError, "error on comparing git sync state"
```
chandankumar4, line 258: should we log the actual `err` if we are returning the custom error message?

juliev0: yeah, I think it would be good to have the original error

juliev0 (approved):

PR #161, opened 2024-03-19
Add configuration for enabling auto heal

### LINKED ISSUE #73: feat: Auto healing
Monitor the live cluster state and auto healing when the live cluster state deviates from git state.

xdevxy:
Fixes #73

### Modifications

This PR adds the configuration for auto heal. By default, the auto heal is enabled with configurable intervals to verifying the cluster states match with the git state.

### Verification

Unit test, also verified in a local cluster.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -92,12 +92,23 @@ func main() {
 		logger.Fatalw("Failed to load config file", zap.Error(err))
 	}
 
+	config, err := configManager.GetConfig()
+	if err != nil {
+		logger.Fatalw("Failed to get config", zap.Error(err))
+	}
+	interval := config.AutoHealTimeIntervalMs
+	// If auto healing is not enabled, use the automated syncing
+	// interval instead.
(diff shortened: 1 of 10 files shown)


cmd/main.go
```diff
@@ -92,12 +92,23 @@ func main() {
...
+	config, err := configManager.GetConfig()
+	if err != nil {
+		logger.Fatalw("Failed to get config", zap.Error(err))
+	}
+	var interval int
+	if config.AutoHeal {
+		interval = config.AutoHealTimeIntervalMs
+	} else {
+		interval = config.SyncTimeIntervalMs
+	}
```
juliev0, line 104: the `AutoHeal` parameter is a little confusing - if it's false, it seems like it should mean "don't auto-heal", but it actually means "use the default auto heal time interval". Perhaps we could just have the `AutoHealTimeIntervalMs` field, and if it's not set, it's essentially the "zero" value, i.e. 0, which means use the default?

PR #163, opened 2024-03-20
refactor: Use the helm and Kustomize pkg from argo-cd

### LINKED ISSUE #87: Use kustomize and helm package from argo-cd
**What happened**:
- Due to client-go version conflict issue with `kubernetes/patcher.go` and argo-cd kustomize package, we have copied the argo-cd kustomize code in numaplane

**What you expected to happen**:
- Use kustomize package from [argo-cd](https://github.com/argoproj/argo-cd/tree/master/util/kustomize)
- Use Helm package from [argo-cd](https://github.com/argoproj/argo-cd/tree/master/util/helm)
- Remove the copied Kustomize and helm code from numaplane

**Anything else we need to know?**:

chandankumar4:
Fixes #87 

### Modifications

- Use helm and kustomize package from argo-cd for generating manifests
- Remove copied code of argod-cd for helm and kustomize

### Verification

Verified on local kind cluster with scenarios as below
- With helm dependency chart
- With remote template in kustomize


internal/git/util.go
```diff
@@ -51,92 +52,129 @@ func GetLatestManifests(
...
+		return "", nil, fmt.Errorf("failed to build manifests, err: %v", err)
+	}
+
+	manifests := make([]string, 0)
+	for _, obj := range targetObjs {
+		if obj == nil {
+			continue
 		}
-		manifestData, err := SplitYAMLToString([]byte(generatedManifest))
+		manifestStr, err := json.Marshal(obj.Object)
```
xdevxy, line 102: If we need to marshal here, we can actually change the signature of the method to return `[]*unstructured.Unstructured` instead so that in `toUnstructuredAndApplyAnnotation` we don't need `Unmarshal` again. HDYT?

xdevxy (commented):
Overall LGTM, just one comment

internal/git/util.go
```diff
@@ -51,92 +52,129 @@ func GetLatestManifests(
...
+		false, "", "", false)
+	if err != nil {
+		return nil, fmt.Errorf("error initializing helm app object: %w", err)
+	}
+
+	templateOpts := &helm.TemplateOpts{
+		Name:      gitSync.Name,
+		Namespace: gitSync.Spec.Destination.Namespace,
+		Set:       map[string]string{},
+		SkipCrds:  true,
```
xdevxy, line 123: What is this config for?

xdevxy (approved):
Approved as [this](https://github.com/numaproj-labs/numaplane/pull/163#discussion_r1532533747) can be a follow up

internal/git/util.go
chandankumar4: This config is for generating the helm manifest by calling method from argo-cd helm package
`out, err := h.Template(templateOpts)`

xdevxy: Sorry I should be specific, I mean `SkipCrds` this is hardcoded as `true` now.

PR #165, opened 2024-03-20
feat: implement clientset

dpadhiar:
### Modifications

Utilizes Kubernetes code-generator to generate clientset files for Numaplane including GitSync client to be used in e2e testing.

This follows the same structure Numaflow uses to construct these files.
Numaflow clientset: https://github.com/numaproj/numaflow/blob/d50c5313a07f1218a3200543676da93ec4b30d9b/pkg/client/clientset/versioned/clientset.go
Script: https://github.com/numaproj/numaflow/blob/main/hack/update-codegen.sh


dpadhiar:
@whynowy @vigith

internal/helm/helm.go
```diff
@@ -0,0 +1,81 @@
...
+	"regexp"
+	"strings"
+
+	numaExec "github.com/numaproj-labs/numaplane/internal/util/exec"
+	"github.com/numaproj-labs/numaplane/pkg/apis/numaplane/v1alpha1"
+)
+
+var (
+	re = regexp.MustCompile(`([^\\]),`)
+)
```
juliev0, line 14: Is this PR up to date with the latest main branch? I don't think this helm change should be showing up here, right?

dpadhiar: I did a pull with main and there was a conflict, this might have gotten mixed up in this. I can revert that to whatever is actually on the main branch.

dpadhiar: I see now this folder wasn't there before, not sure how that got added. Will remove.

hack/update-codegen.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
xdevxy, line 1: These files are copied from numaflow, right?

xdevxy (approved):

Issue #144, opened 2024-03-11
Docker test image for GitServer 


juliev0:
# Summary

Create a Docker image to run Git Server to be used for unit test, e2e test, and potentially BDD test for AddOn, which can optionally mount a volume.  It can initialize 0-n repositories based on command line input. 

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

xdevxy:
Closed by #169

PR #169, opened 2024-03-21
local git server with kubernetes

shubhamdixit863:
Fixes (https://github.com/numaproj-labs/numaplane/issues/144)

### Modifications

A local Git server running behind an Apache HTTP server is created using a Dockerfile. The Git server is served through Apache using the git-httpd CGI module. The Dockerfile creates a variable number of repositories depending on the BUILD ARG passed; a Makefile script has been added for the same purpose. Moreover, a gitserver.yaml file, creating a StatefulSet and a service to access the Git server running on a Kubernetes pod, has also been created. The required authorized_key file for SSH and .httpsswd for Apache are mounted by the config map generated by Kustomize.

Edit 
An api to create the remote repo has also been added ,now you can create the repos once the container is started using http api -
#curl -X POST -d "name=my_new" http://localhost:8080/create Response {"status": "success", "message": "Repository 'my_new' created successfully."}

### Verification

Everything has been tested by running the  `kustomize` from `tests/e2e-gitserver` using command` kubectl apply -k . `     in  local Kubernetes cluster.


config/samples/numaplane.numaproj.io_v1alpha1_gitsync.yaml
```diff
@@ -4,8 +4,8 @@ metadata:
   name: gitsync-example
   namespace: numaplane-system
 spec:
-  path: sample-pipeline
-  repoUrl: https://github.com/xdevxy/numaflow-example-pipelines.git
+  path: ""
+  repoUrl: https://github.com/shubhamdixit863/yamlrepo
```
juliev0, line 8: let's try not to include any personal forks

juliev0:
So, in the original issue, I was assuming the start.sh file would process the incoming arguments and use that to determine which repos (or at least the number of repos) need a "git init" and then do it there. As a build arg, it looks like we would basically need a different image for each count? Maybe it's no big deal if there are 5 repos by default, assuming there's volume mounting. But if I want to use the Dockerfile with no volume mounted and with 0 repositories, I need a separate image for that. 

Let me think about this, whether we need what I was originally looking for or not.

juliev0:
Updated previous comment fyi, in case you missed it...what do you think of what I was saying above? I think if it unblocks Dillen, we could merge it the way you have it, but then modify to what I mentioned above, if it's not too difficult.

I guess we can think about what would be needed for this to work for e2e test, unit test, and BDD test.

Assuming there's always a user for all of these, then theoretically the user can always run the "git init", etc. commands I believe, right?

juliev0:
> Assuming there's always a user for all of these, then theoretically the user can always run the "git init", etc. commands I believe, right?

I spoke to Shubham offline. Apparently, my statement here isn't true, as the server side needs to set up the git repository to begin with, tied to the directory location on the server where it lives. 

Shubham will investigate seeing if he can get it working to move the git init logic from build time to run time, however. (If there are any significant challenges, however, we can merge this as is, and not worry about the potential volume mounting.)

tests/e2e-gitserver/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+RUN for i in $(seq 1 $REPO_COUNT); do \
+    mkdir /tmp/git-repo$i && \
+    cd /tmp/git-repo$i && \
+    git init && \
+    git config --global user.email "you@example.com" && \
+    git config --global user.name "Your Name" && \
+    echo "some: data" > data.yaml && \
+    git add data.yaml && \
+    git commit -m "Initial commit" && \
+    git push /var/www/git/repo$i.git master;\
```
juliev0, line 89: does this mean that the directory `/var/www/git/repo$i.git` is considered to be the _remote_?

shubhamdixit863: yes its treated as remote . And is being server through apache http

Makefile
```diff
@@ -193,3 +197,13 @@ $(CONTROLLER_GEN): $(LOCALBIN)
...
+
+.PHONY: gitserver
+gitserver:
+		cd tests/e2e-gitserver && \
+		 docker buildx build \
+		--no-cache \
+		--build-arg REPO_COUNT=$(REPO_COUNT) \
+		--platform $(PLATFORMS) \
+		-t quay.io/numaio/$(GITSERVER_NAME):latest \
+		--push .
```
juliev0, line 209: Since IMAGE_NAMESPACE is [configurable](https://github.com/numaproj-labs/numaplane/pull/169/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L13), maybe this should be as well?

juliev0:
can we move the Dockerfile and its dependencies out of the e2e directory, since it can be used for the unit test and BDD test as well?

config/samples/numaplane.numaproj.io_v1alpha1_gitsync.yaml
juliev0: can you fix this?

Makefile
```diff
@@ -31,6 +31,10 @@ GCFLAGS="all=-N -l"
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.28.0
 
+# LOCAL GIT SERVER CONFIG
+REPO_COUNT=5
+GITSERVER_NAME=quay.io/numaio/numaplane-e2e-gitserver:latest
```
juliev0, line 36: nit: can we call it `GITSERVER_IMAGE`?

tests/e2e-gitserver/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+#kubectl apply -k .
```
juliev0, line 1:
```suggestion
```

tests/gitserver/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+# Enable the site and ensure the Apache user can read the .htpasswd file
+RUN a2ensite 000-default && \
+    chmod 644 /auth/.htpasswd && \
+    chown www-data:www-data /auth/.htpasswd
+
+# Expose port 22 for SSH access, 80 for HTTP access, and 443 for HTTPS access
+EXPOSE 22 80 443
+
+
+RUN for i in $(seq 1 $REPO_COUNT); do \
```
juliev0, line 86: add comment for this section

Makefile
```diff
@@ -31,6 +31,10 @@ GCFLAGS="all=-N -l"
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.28.0
 
+# LOCAL GIT SERVER CONFIG
+REPO_COUNT=5
+GITSERVER_IMAGE=quay.io/numaio/numaplane-e2e-gitserver:latest
```
juliev0, line 36:
```suggestion
GIT_SERVER_VERSION ?= latest
GITSERVER_IMAGE=quay.io/numaio/numaplane-e2e-gitserver:$(GIT_SERVER_VERSION)
```

juliev0 (approved):

PR #172, opened 2024-03-22
fix: broken Dockerfile, missing pkg path after api dir was moved/dele…

afugazzotto:
This PR solves the following bugs:
- docker build was failing due to missing `api` dir since it was moved under `pkg/apis`
- upgraded `controller-gen` from `v0.13.0` to `v0.14.0` to resolve tests failure on local env

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,7 +9,7 @@ issues:
   # restore some of the defaults
   # (fill in the rest as needed)
   exclude-rules:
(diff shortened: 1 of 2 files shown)


dpadhiar (commented):
Resolved errors from moving `api` directory to `pkg` in my previous PR.

juliev0 (approved):

PR #175, opened 2024-03-25
feat: implement given/when/expect structs for e2e testing

### LINKED ISSUE #162: create Given / When / Expect structs

dpadhiar:
Fixes #162 

### Modifications

Implements given/when/expect structs for e2e testing.

Given constructs GitSync CRDs provided by test or file in given path.
When creates and deletes GitSyncs in k8s as well as pushing Git commits to git-server pod to test functionality.
Expect verifies that resource are being created and deleted in target namespace.


juliev0:
Have you been able to test anything given that it's not being called from anywhere yet?

dpadhiar:
> Have you been able to test anything given that it's not being called from anywhere yet?

Not yet but I could give it a try tomorrow after standup by writing some basic calls.

tests/fixtures/expect.go
```diff
@@ -0,0 +1,75 @@
...
+	return e
+}
+
+// check that resources are deleted
+func (e *Expect) ResourcesDontExist(resourceType string, resources []string) *Expect {
+
+	e.t.Helper()
+	ctx := context.Background()
+	for _, r := range resources {
+		result := e.kubeClient.CoreV1().RESTClient().Get().Timeout(defaultTimeout).Namespace(TargetNamespace).Resource(resourceType).Name(r).Do(ctx)
```
juliev0, line 58: Is this line right? It looks like it might be the same as the function above? (I'm not sure what `Do()` is)

dpadhiar: Do outputs the result that we check for the Get request. It is the same logic but we're looking for an error to tell us the resources don't exist. This might not be the best approach. Alternatively, we could list Pods and check that those don't exist.

tests/fixtures/given.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (g *Given) WithGitSync(gs *v1alpha1.GitSync) *Given {
+	g.t.Helper()
+	g.gitSync = gs
+	l := g.gitSync.GetLabels()
+	if l == nil {
+		l = map[string]string{}
+	}
+	l[Label] = LabelValue
+	g.gitSync.SetLabels(l)
```
juliev0, line 62: nit: can we create a function for this since it's duplicate code from above?

tests/fixtures/expect.go
```diff
@@ -0,0 +1,75 @@
...
+		if result.Error() != nil {
```
dpadhiar, line 59: Ah, this should be == nil. Will fix.

tests/fixtures/when.go
```diff
@@ -0,0 +1,165 @@
...
+	for {
+		select {
+		case event := <-watch.ResultChan():
+			i, ok := event.Object.(*v1alpha1.GitSync)
+			if ok {
+				if i.Status.Phase == v1alpha1.GitSyncPhaseRunning {
+					return nil
+				}
+			} else {
+				return fmt.Errorf("not gitsync")
```
juliev0, line 159: could be GitSyncPhasePending, which means it's about to become GitSyncPhaseRunning, which wouldn't be an error

dpadhiar: Will make this change then: 
```
				if i.Status.Phase == v1alpha1.GitSyncPhaseRunning || i.Status.Phase == v1alpha1.GitSyncPhasePending {
```

tests/fixtures/when.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// open worktree
+	wt, err := repo.Worktree()
+	if err != nil {
+		w.t.Fatal(err)
+	}
+
+	// iterate over files to be added and committed
+	for _, f := range files {
+		_, err = wt.Add(f)
```
juliev0, line 95: can this handle files that are subpaths within the repository path?

tests/fixtures/when.go
```diff
@@ -0,0 +1,165 @@
...
+	ctx := context.Background()
+	if err := w.gitSyncClient.Delete(ctx, w.gitSync.Name, metav1.DeleteOptions{}); err != nil {
+		w.t.Fatal(err)
+	}
+
+	return w
+}
+
+// make git push to Git server pod
+func (w *When) PushToGitRepo(repoPath string, files []string) *When {
```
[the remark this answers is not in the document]
dpadhiar: That would make sense, I was thinking of cases where controlling what repo the commit is sent to would be useful but I don't have any as of now.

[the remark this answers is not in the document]
juliev0: Keep in mind also that a `GitSync` defines [both](https://github.com/numaproj-labs/numaplane/blob/main/pkg/apis/numaplane/v1alpha1/shared_types.go#L65-L71) a `RepoURL` and a `Path` within that repo

tests/fixtures/when.go
dpadhiar: As in adding `dir` as a whole or `dir/file.yaml`? I believe it supports both but I can check the documentation.

https://github.com/go-git/go-git/blob/v5.11.0/worktree_status.go#L272

If a directory given, adds the files and all his sub-directories recursively in the worktree to the index.

tests/fixtures/when.go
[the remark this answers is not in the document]
dpadhiar: Out of those, we'd want the RepoURL.

tests/fixtures/expect.go
juliev0: I think it won't work with the timeout you're passing. It will check if everything exists, which it initially may if everything's just been deleted.

tests/fixtures/when.go
juliev0: Can you wait until it's Running if it's still Pending?

tests/fixtures/when.go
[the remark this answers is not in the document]
dpadhiar: > so, I'm not sure if you want the files to be full paths or paths within the GitSync.Path?

Looking at it now, we need `repoURL` to get the working tree. I'm undecided on what we need for `files` between full paths or paths within the `GitSync.Path`. If the path is `sample-pipelines` for example, we would just want to commit that.

tests/fixtures/expect.go
```diff
@@ -0,0 +1,85 @@
...
+
+	e.t.Helper()
+	ctx := context.Background()
+	for _, r := range resources {
+		result := e.kubeClient.CoreV1().RESTClient().Get().
+			Timeout(defaultTimeout).
+			Namespace(TargetNamespace).
+			Resource(resourceType).
+			Name(r).
+			Do(ctx)
```
juliev0, line 68: I don't think this will work. Let's say the call to the K8S API to delete the resources was just called, and then this is called. This will find the resource which hasn't quite been deleted yet and it will return an error. The `defaultTimeout` won't have done its job.

tests/fixtures/given.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (g *Given) WithGitSync(gs *v1alpha1.GitSync) *Given {
+	g.t.Helper()
+	g.gitSync = gs
+	g.addLabel()
+	return g
+}
+
+func (g *Given) addLabel() {
```
juliev0, line 56: nit: let's rename to `addE2ELabel()` since that's what it is (right now sounds like a generic label)

tests/fixtures/e2e_suite.go
```diff
@@ -33,9 +35,11 @@ import (
...
 const (
 	/* resource names */
-	Namespace      = "numaplane-system"
-	Label          = "numaplane-e2e"
-	defaultTimeout = 60 * time.Second
+	Namespace       = "numaplane-system"
+	TargetNamespace = "numaflow-pipeline"
+	Label           = "numaplane-e2e"
+	LabelValue      = "true"
+	defaultTimeout  = 60 * time.Second
```
juliev0, line 42: nit: rename `Label` to `E2ELabel` and rename `LabelValue` to `E2ELabelValue`

tests/fixtures/when.go
```diff
@@ -0,0 +1,171 @@
...
+	w.t.Helper()
+	if w.gitSync == nil {
+		w.t.Fatal("No gitsync to delete")
+	}
+	w.t.Log("Deleting gitsync", w.gitSync.Name)
+	ctx := context.Background()
+	if err := w.gitSyncClient.Delete(ctx, w.gitSync.Name, metav1.DeleteOptions{}); err != nil {
+		w.t.Fatal(err)
+	}
+
```
juliev0, line 73: nit: wonder if we might have symmetry between `CreateGitSyncAndWait()` and this method? If that one waits, is there any reason that this one shouldn't wait as well?

tests/fixtures/when.go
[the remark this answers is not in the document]
juliev0: that's a good question. will let you decide

tests/fixtures/when.go
```diff
@@ -0,0 +1,171 @@
...
+		timeoutCh <- true
+	}()
+	for {
+		select {
+		case event := <-watch.ResultChan():
+			i, ok := event.Object.(*v1alpha1.GitSync)
+			if ok {
+				// gitSync is about to start
+				if i.Status.Phase == v1alpha1.GitSyncPhasePending {
+					continue
```
juliev0, line 159: can we log that it's in pending? (which could help diagnose if it every stays in pending)

tests/fixtures/when.go
dpadhiar: We can achieve this similar to how Numaflow checks that a Pipeline is deleted by listing all Pods for the Pipeline and checking that the returned list size is 0.

For a GitSync, we can use the client to list them and check that this list is 0 as well.

juliev0: We just need to check that the GitSync has been deleted

dpadhiar: We could wait for gitSync.Get() to fail.

juliev0: Sure. I think ideally also check the error to see if it's one which indicates that the resource doesn't exist, as opposed to another failure.

juliev0: https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/controller.go#L567
`apiError.IsNotFound()`

dpadhiar: Sounds good, this is the current implementation now for checking GitSync and resources:

```
if err != nil {
			if errors.IsNotFound(err) {
				return w
			} else {
				w.t.Fatalf("Error getting gitSync: %v", err)
			}
		}
```

tests/fixtures/expect.go
```diff
@@ -0,0 +1,107 @@
...
+		result := e.kubeClient.CoreV1().RESTClient().Get().
+			Namespace(TargetNamespace).
+			Resource(resourceType).
+			Name(resource).
+			Do(ctx)
+		if result.Error() != nil {
+			if errors.IsNotFound(result.Error()) {
+				return true
+			} else {
+				return false
```
juliev0, line 101: This is a nice way to solve the timeout issue.

I would say that in the else case, we can consider whether to return false or not. If it's a temporary network error, it would be okay and we could continue....but perhaps that network error would be rare running locally anyway. At the very least, let's log something to distinguish between this and the other "return false".

dpadhiar: I'm not sure yet how we would determine which network error would be okay to continue on, for now we can just log the error to make it clearer. I can also log a different error for the other false.

juliev0: that sounds good

tests/fixtures/when.go
```diff
@@ -0,0 +1,197 @@
...
+		time.Sleep(2 * time.Second)
+	}
+}
+
+// make git push to Git server pod
+func (w *When) PushToGitRepo(files []string) *When {
+
+	// open path to git server
+	// an example path would be http://localhost:8080/git/repo1.git
+	repo, err := git.PlainOpen(w.gitSync.Spec.RepoUrl)
```
juliev0, line 106: Will you need to have a separate function to clone the repo at first before you do this? See [this](https://github.com/numaproj-labs/numaplane/blob/main/internal/git/util.go#L261-L268) code. It attempts to clone and then just opens if it's already been cloned. You could do something similar or have a separate function to do the clone first.

Note this functionality is cloning to disk, but I guess there's some alternative option in go-git in which the clone can be to memory instead. You can do either, but if it's disk, we should try to make sure the files get cleaned up and are in a .gitignore.

tests/fixtures/when.go
juliev0: thanks. Hopefully it won't log too many messages. If it seems bad for some reason, we could revisit :)

tests/fixtures/when.go
```diff
@@ -0,0 +1,197 @@
...
+	if w.gitSync == nil {
+		w.t.Fatal("No gitsync to delete")
+	}
+	w.t.Log("Deleting gitsync", w.gitSync.Name)
+	ctx := context.Background()
+	if err := w.gitSyncClient.Delete(ctx, w.gitSync.Name, metav1.DeleteOptions{}); err != nil {
+		w.t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
```
juliev0, line 81: Since you have `waitForGitSyncRunning()` above, maybe we should have `waitForGitSyncDeleted()` for symmetry?

tests/fixtures/when.go
dpadhiar: I'll need to test how this works but that seems to be the right approach based on that other code.

I am thinking about how we will need to clone the repo to make sure we can commit all our test files easily.

tests/fixtures/when.go
```diff
@@ -0,0 +1,228 @@
...
+		_, err := w.gitSyncClient.Get(ctx, w.gitSync.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		time.Sleep(2 * time.Second)
+	}
+
+}
+
+func (w *When) cloneRepo(ctx context.Context) (*git.Repository, error) {
```
juliev0, line 211:
```suggestion
// clone repository unless it's already been cloned
func (w *When) cloneRepo(ctx context.Context) (*git.Repository, error) {
```

tests/fixtures/given.go
```diff
@@ -0,0 +1,188 @@
...
+		g.t.Fatal(err)
+	}
+
+	return g
+}
+
+// clone repository unless it's already been cloned
+func (g *Given) cloneRepo(ctx context.Context) (*git.Repository, error) {
+
+	path := "./tmp"
```
juliev0, line 163: nit: let's not list this path twice in the code...maybe a variable for it or something

tests/fixtures/given.go
```diff
@@ -0,0 +1,188 @@
...
+	repo, err := g.cloneRepo(ctx)
+	if err != nil {
+		g.t.Fatal(err)
+	}
+
+	tmpPath := filepath.Join("tmp", g.gitSync.Spec.Path)
+	dataPath := filepath.Join("testdata", g.gitSync.Spec.Path)
+
+	// create dir for gitsync path
+	_ = os.Mkdir(tmpPath, 0777)
```
juliev0, line 118: doesn't it already exist here because you added it when you called `cloneRepo()`?

juliev0: never mind, it's the entire path...sorry

tests/fixtures/given.go
```diff
@@ -0,0 +1,188 @@
...
+
+	dir, err := os.ReadDir(dataPath)
+	if err != nil {
+		g.t.Fatal(err)
+	}
+
+	// assumes no nested directory for now
```
juliev0, line 125: in that case should we make this "todo" to add nested capability?

dpadhiar: Will add a comment for that.

tests/fixtures/given.go
```diff
@@ -0,0 +1,196 @@
...
+
+	cloneOpts := git.CloneOptions{URL: g.gitSync.Spec.RepoUrl, Auth: auth}
+
+	repo, err := git.PlainCloneContext(ctx, tempPath, false, &cloneOpts)
+	if err != nil && errors.Is(err, git.ErrRepositoryAlreadyExists) {
+		existingRepo, openErr := git.PlainOpen(tempPath)
+		if openErr != nil {
+			return repo, fmt.Errorf("failed to open existing repo: %v", openErr)
+		}
+		return existingRepo, nil
```
juliev0, line 181: later consider if you really need this backup method of using the existing or not....i.e. if this only should get called when you are cloning for the first time...

tests/fixtures/expect.go
```diff
@@ -0,0 +1,109 @@
...
+type Expect struct {
+	t             *testing.T
+	restConfig    *rest.Config
+	kubeClient    kubernetes.Interface
+	gitSync       *v1alpha1.GitSync
+	gitSyncClient planepkg.GitSyncInterface
+}
+
+// check that resources are created
+func (e *Expect) ResourcesExist(resourceType string, resources []string) *Expect {
```
juliev0, line 40: later consider if this is the function signature you want, or perhaps a `map[Resource]` where a `Resource` is a struct consisting of resourceType and name or something....fine for now, though

juliev0 (approved):

Issue #137, opened 2024-03-07
Ensure delete a GitSync CR is handled correctly


xdevxy:
# Summary

When an existing GitSync CR gets delete, we need to all resources belongs to it are deleted together. This means when we an GitSync is deleted, we need to delete all the resources associated with it.

Particularly, we need to delete the resources at:
- during controller [reconciliation](https://github.com/numaproj-labs/numaplane/blob/main/internal/controller/gitsync_controller.go#L140), which means the GitSync may no longer be enqueued.
- during the [worker task execution](https://github.com/numaproj-labs/numaplane/blob/main/internal/sync/syncer.go#L210)
- after the worker task [finishes syncing](https://github.com/numaproj-labs/numaplane/blob/main/internal/sync/syncer.go#L242)

This ensures with a deleted gitsync after the last sync, all associated resources are deleted as the gitsync will not be enqueued again.

To retrieve the resources associated with the GitSync, we should utilize `LiveStateCache.GetManagedLiveObjs()`. Moreover, we may want to add an option in configmap to enable or disable cascading deletion.
 
Note that if `Not Found` error occur during deleting the resources, it should be ignored.

# Use Cases

Delete existing GitSync CRs

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
We need to come up with a good way to do this so as to not have any race conditions actually: if we do it in the reconciler when we notice the `DeletionTimestamp` [here](https://github.com/numaproj-labs/numaplane/blob/4349bca446c647882ec321201697ca609b7bacb6/internal/controller/gitsync_controller.go#L140), then what if at the same time we're basically at [this](https://github.com/numaproj-labs/numaplane/blob/4349bca446c647882ec321201697ca609b7bacb6/internal/sync/syncer.go#L221) line performing the sync?

juliev0:
Okay, @xdevxy and I discussed that the `DeletionTimestamp` can be checked again after the sync to perform deletion there as well just in case that race condition occurs.

xdevxy:
Yeah, since the gitsync will no longer be enqueued again and we just need to ensure after the last sync the resources are deleted in case it gets there.

PR #176, opened 2024-03-26
Delete gitsync's Child Resources In the Cluster If Gitsync Gets Deleted

shubhamdixit863:
Yes, it fixes [Issue #137](https://github.com/numaproj-labs/numaplane/issues/137).

### Modifications

[DeleteManagedObjectsGitSync](https://github.com/numaproj-labs/numaplane/blob/deletGitsync/internal/util/kubernetes/kubernetes.go#L113) has been added, which is called at the places mentioned in [the aforementioned issue](https://github.com/numaproj-labs/numaplane/issues/137). This method is designed to get all the resources from Syncer Cache  and subsequently delete them.
Moreover a field  in Global Config Has been added to enable or disable the cascade deletion of the resources.

### Verification

Unit tests have been written for the `DeleteManagedObjectsGitSync` method, as documented [here](https://github.com/numaproj-labs/numaplane/blob/deletGitsync/internal/util/kubernetes/kubernetes_test.go#L173). Furthermore, the deletion functionality has been tested by running the controller and Gitsync on a local Kubernetes cluster.
If gitsync is deleted ,then the related resources are deleted as well having the same annotation key value pair


internal/util/kubernetes/kubernetes.go
```diff
@@ -86,10 +90,44 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
...
+
+// DeleteResourcesByAnnotations deletes all resources of a given GroupVersionKind across the kubernetes cluster
+func DeleteResourcesByAnnotations(ctx context.Context, client k8sClient.Client, gvks []schema.GroupVersionKind, annotationKey string, annotationValue string) error {
+	logger := logging.FromContext(ctx)
+	for _, gvk := range gvks {
+		listItems := &unstructured.UnstructuredList{}
+		listItems.SetGroupVersionKind(gvk)
+
+		// List all resources of the given GroupVersionKind across the cluster
+		if err := client.List(ctx, listItems); err != nil {
```
juliev0, line 116: this will eventually be a very large list of items, so a preferred approach would be to use a Label rather than an Annotation, where a Label actually allows you to use a selector to select by that label.

internal/common/common.go
```diff
@@ -10,3 +12,13 @@ const (
...
 )
+
+// PredefinedGroupVersionKinds contains the list of GVK which we need to delete in the cluster post gitsync gets deleted
+var PredefinedGroupVersionKinds = []schema.GroupVersionKind{
+	{Group: "", Version: "v1", Kind: "Pod"},
+	{Group: "apps", Version: "v1", Kind: "Deployment"},
+	{Group: "apps", Version: "v1", Kind: "ReplicaSet"},
+	{Group: "apps", Version: "v1", Kind: "StatefulSet"},
+	{Group: "apps", Version: "numaflow.numaproj.io/v1alpha1", Kind: "Pipeline"},
+	{Group: "apps", Version: "numaflow.numaproj.io/v1alpha1", Kind: "InterStepBufferService"},
```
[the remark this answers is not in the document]
juliev0: @xdevxy does the gitops-engine framework require the use of an Annotation or can we use a Label instead? (or we can use both if not)

internal/util/kubernetes/kubernetes.go
juliev0: @xdevxy does the gitops-engine framework require the use of an Annotation or can we use a Label instead? (or we can use both if not)

xdevxy: Oh, apologize that I should mention in the issue that we should utilize the cluster cache. Specifically, [LiveStateCache.GetManagedLiveObjs()](https://github.com/numaproj-labs/numaplane/blob/main/internal/sync/cache.go#L79rl) which you can get all the live objs associated to the given GitSync. I'll update the issue as well.

internal/sync/syncer.go
```diff
@@ -239,6 +245,18 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
 		Namespace: gitSync.Namespace,
 		Name:      gitSync.Name,
 	}
+
+	if !gitSync.GetDeletionTimestamp().IsZero() {
```
xdevxy, line 249: This should be the last thing during sync task and after we do `updateCommitStatus`.

internal/controller/gitsync_controller.go
```diff
@@ -137,6 +138,12 @@ func (r *GitSyncReconciler) reconcile(ctx context.Context, gitSync *apiv1.GitSyn
 	if !gitSync.DeletionTimestamp.IsZero() {
 		logger.Infow("Deleting", "GitSync", gitSync)
 		if r.syncer != nil {
+			// Delete the linked resources to the GitSync
+			// Is gitSync.Name is the correct value for annotation ?
+			err := kubernetesshared.DeleteResourcesByAnnotations(ctx, r.client, common.PredefinedGroupVersionKinds, common.AnnotationKeyGitSyncInstance, gitSync.Name)
```
xdevxy, line 143: At here and all other places we delete the resources, we should not returning the error if it is a `NotFound` error. This is stated in the issue description: "Note that if Not Found error occur during deleting the resources, it should be ignored."

xdevxy (commented):
@shubhamdixit863 Other than the comments I left,  can you add an option in ConfigMap to enable or disable this cascade deletion feature. By default, let's put it as disabled. Thanks!

internal/controller/gitsync_controller.go
shubhamdixit863: I am already ignoring that in here -https://github.com/numaproj-labs/numaplane/blob/deletGitsync/internal/util/kubernetes/kubernetes.go#L117

internal/common/common.go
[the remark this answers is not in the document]
shubhamdixit863: Post Live managed objects its not required now

shubhamdixit863:
> @shubhamdixit863 Other than the comments I left, can you add an option in ConfigMap to enable or disable this cascade deletion feature. By default, let's put it as disabled. Thanks!

You mean add in the config.yaml [here](https://github.com/numaproj-labs/numaplane/blob/deletGitsync/config/samples/config_map.yaml#L8)   then I need to add a global config variable too [here](https://github.com/numaproj-labs/numaplane/blob/deletGitsync/internal/controller/config/config.go#L35)

xdevxy:
> > @shubhamdixit863 Other than the comments I left, can you add an option in ConfigMap to enable or disable this cascade deletion feature. By default, let's put it as disabled. Thanks!
> 
> You mean add in the config.yaml [here](https://github.com/numaproj-labs/numaplane/blob/deletGitsync/config/samples/config_map.yaml#L8) then I need to add a global config variable too [here](https://github.com/numaproj-labs/numaplane/blob/deletGitsync/internal/controller/config/config.go#L35) . As until its not available in GlobalConfig how will I get it as a variable .

Yes, you need to do that

internal/controller/gitsync_controller.go
shubhamdixit863: I have made the changes  as per Syncer Cache so thats why its not working .Not relevant now .Thanks

internal/sync/syncer.go
```diff
@@ -206,8 +207,21 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
...
+			return err
+		}
+		// if cascade deletion is enabled in the config , Delete the linked resources to the GitSync
+		if config.CascadeDeletion {
+			err := CascadeDeletion(ctx, s.client, s.stateCache, gitSync)
+			if err != nil {
+				return err
+			}
+		}
+		s.StopWatching(key)
```
juliev0, line 224: all of this code is copy/pasted 3 times. Let's create a method called `ProcessGitSyncDeletion()` or something

internal/util/kubernetes/kubernetes.go
```diff
@@ -86,10 +88,28 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
...
+
 func IsValidKubernetesManifestFile(fileName string) bool {
 	fileExt := strings.Split(fileName, ".")
 	if _, ok := validManifestExtensions[fileExt[len(fileExt)-1]]; ok {
 		return true
 	}
 	return false
 }
+
+// DeleteManagedObjectsGitSync deletes all resources of a given GroupVersionKind across the kubernetes cluster
```
juliev0, line 106: I think this is an outdated comment

internal/util/kubernetes/kubernetes_test.go
```diff
@@ -154,3 +159,55 @@ func TestIsValidKubernetesManifestFile(t *testing.T) {
...
+	err = fakeClient.Create(ctx, resource)
+	assert.NoError(t, err)
+
+	objs := map[kube.ResourceKey]*unstructured.Unstructured{
+		{
+			Name:      "test-name",
+			Namespace: "test-namespace",
+		}: resource,
+	}
+	// Now call DeleteResourcesByAnnotations
```
juliev0, line 203: outdated comment

internal/util/kubernetes/kubernetes_test.go
```diff
@@ -154,3 +159,55 @@ func TestIsValidKubernetesManifestFile(t *testing.T) {
...
+			Name:      "test-pod",
+			Namespace: "numaplane-test",
+		},
+	}
+	fakeClient := fake.NewClientBuilder().WithScheme(scheme).WithObjects(pod).Build()
+	err = DeleteKubernetesResource(context.Background(), fakeClient, pod)
+	assert.NoError(t, err)
+}
+
+func TestDeleteResourcesByAnnotation(t *testing.T) {
```
juliev0, line 178: outdated function name

internal/sync/syncer.go
```diff
@@ -239,7 +253,31 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
 		Namespace: gitSync.Namespace,
 		Name:      gitSync.Name,
 	}
-	return updateCommitStatus(ctx, s.client, log, namespacedName, commitHash, synced, syncMessage)
+
+	err = updateCommitStatus(ctx, s.client, log, namespacedName, commitHash, synced, syncMessage)
+	if err != nil {
```
xdevxy, line 258: Apologize on this,  but even with errors we should check if the GitSync is deleted. Thus, I think the code should be,
```
if !gitSync.GetDeletionTimestamp().IsZero() {
   ...
} else {
   return updateCommitStatus(ctx, s.client, log, namespacedName, commitHash, synced, syncMessage)
}
```

internal/util/kubernetes/kubernetes.go
```diff
@@ -86,10 +88,28 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
...
+func DeleteManagedObjectsGitSync(ctx context.Context, client k8sClient.Client, objs map[kube.ResourceKey]*unstructured.Unstructured) error {
+	for _, obj := range objs {
+		if err := DeleteKubernetesResource(ctx, client, obj); err != nil {
```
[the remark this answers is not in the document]
shubhamdixit863: I am  catching and ignoring the error here -https://github.com/numaproj-labs/numaplane/blob/deletGitsync/internal/util/kubernetes/kubernetes.go#L96 seems better

internal/sync/syncer.go
```diff
@@ -393,3 +409,51 @@ func updateCommitStatus(
...
+}
+
+func ProcessGitSyncDeletion(ctx context.Context, gitSync *v1alpha1.GitSync, s *Syncer) error {
+	config, err := controllerConfig.GetConfigManagerInstance().GetConfig()
+	if err != nil {
+		return err
+	}
+	// if cascade deletion is enabled in the config ,Delete the linked resources to the GitSync
+	if config.CascadeDeletion {
+		err := CascadeDeletion(ctx, s.client, s.stateCache, gitSync)
```
xdevxy, line 453: nit: `return CascadeDeletion(ctx, s.client, s.stateCache, gitSync)` is good enough here.

internal/sync/syncer.go
```diff
@@ -393,3 +409,51 @@ func updateCommitStatus(
...
 }
+
+// GetStateCache is the exported method for getting the cache  in other packages
+func (s *Syncer) GetStateCache() LiveStateCache {
+	return s.stateCache
+}
+
+// GetLiveManagedObjects retrieves live managed objects from the provided cache for the given GitSync configuration.
+// It returns a map of Kubernetes resource keys to unstructured objects and an error if any.
+func GetLiveManagedObjects(cache LiveStateCache, gitSync *v1alpha1.GitSync) (map[kube.ResourceKey]*unstructured.Unstructured, error) {
```
xdevxy, line 420: nit: Why do we need to add these getter functions and expose it externally?

xdevxy (approved):
Please address the nits in this PR or a followup

internal/util/kubernetes/kubernetes.go
```diff
@@ -86,10 +90,32 @@ func GetSecret(ctx context.Context, client k8sClient.Client, namespace, secretNa
...
 func IsValidKubernetesManifestFile(fileName string) bool {
 	fileExt := strings.Split(fileName, ".")
 	if _, ok := validManifestExtensions[fileExt[len(fileExt)-1]]; ok {
 		return true
 	}
 	return false
 }
+
+// DeleteManagedObjectsGitSync deletes Kubernetes resources from a map sequentially, returning an error if any deletion fails.
+func DeleteManagedObjectsGitSync(ctx context.Context, client k8sClient.Client, objs map[kube.ResourceKey]*unstructured.Unstructured) error {
```
juliev0, line 114: nit: let's change the name of this function because it actually has nothing to do with `GitSync` per se. you're just deleting some resources

internal/sync/syncer.go
shubhamdixit863: yeah it was not required ,removed it.

Issue #127, opened 2024-03-05
chore: consolidate logging into a single library


xdevxy:
Use a single logging library with fine grained levels and valid to use with gitops engine.

whynowy:
`logrus` has stepped down, use `zerolog` or `zap` instead.

PR #177, opened 2024-03-26
Logging

afugazzotto:
Fixes https://github.com/numaproj-labs/numaplane/issues/127

### Modifications

Replaced the 2 logging libraries (zap and logrus) with a single library (zerolog). This decision was made to consolidate multiple logging utilities into one, standardize log output, allow for future features, and possible to allow reusability of this logger with other numaproj tools. Tried to keep older functionality unchanged as much as possible. 

### Verification

Tested the changes manually by inspecting the logs and also wrote unit tests with 87.7% coverage.


juliev0:
if @xdevxy is good with this, I don't need to review as well..thanks!

cmd/agent/main.go
```diff
@@ -20,12 +20,12 @@ import (
...
 
 	"github.com/numaproj-labs/numaplane/internal/agent"
-	"github.com/numaproj-labs/numaplane/internal/util/logging"
+	"github.com/numaproj-labs/numaplane/internal/util/logger"
 )
 
 var (
 	// logger is the global logger for the numaplane-agent
-	logger     = logging.NewLogger().Named("numaplane-agent")
+	numaLogger = logger.New(nil, nil).WithName("numaplane-agent")
```
xdevxy, line 28: As discussed add a new method `NewWithWriter` and keep this one (`New`) without any parameters if we don't need them.

internal/util/logger/logger.go
```diff
@@ -0,0 +1,291 @@
...
+	zerolog.SetGlobalLevel(logrVerbosityToZerologLevelMap[verboseLevel])
+
+	w := io.Writer(os.Stdout)
+	if writer != nil {
+		w = *writer
+	}
+
+	lvl := int(defaultLevel)
+	if level != nil {
+		lvl = *level
```
xdevxy, line 88: I don't follow how this `level` is used? and it is default to debug level if not provided?

afugazzotto: `level` is the argument to the function. If it is omitted, the function will use a default level (I decided to use Info level for default).

internal/common/common.go
```diff
@@ -5,8 +5,6 @@ const (
 	AnnotationKeyGitSyncInstance = "numaplane.numaproj.io/tracking-id"
 	// SSAManager is the default numaplane manager name used by server-side apply syncs
 	SSAManager = "numaplane-controller"
-	// EnvLogFormat log format that is defined by `--logformat` option
-	EnvLogFormat = "NUMAPLANE_LOG_FORMAT"
 	// EnvLogLevel log level that is defined by `--loglevel` option
```
xdevxy, line 8: nit: is this comment still hold true? e.g. we still have`--loglevel` ?

xdevxy (approved):

internal/common/common.go
afugazzotto: I removed the NUMAPLANE_LOG_FORMAT because it no longer applies.

PR #180, opened 2024-03-29
Added log level in confi

afugazzotto:
### Modifications

Added a LogLevel field to global and agent config structs to be able to load the log level from configmap. The log level is updated every time the config is retrieved again in case the level was changed in configmap.

### Verification

Manually tested by inspecting the logs.

### DIFF
--- a/internal/agent/agent_syncer.go
+++ b/internal/agent/agent_syncer.go
@@ -100,6 +100,9 @@ func (syncer *AgentSyncer) checkConfigUpdate() bool {
 			return false
 		}
 		syncer.configRevision = newRevision
+
+		syncer.numaLogger.SetLevel(syncer.config.LogLevel)
+
 		return true
 	}
 	return false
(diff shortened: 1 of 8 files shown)


internal/agent/agent_syncer.go
```diff
@@ -100,6 +100,9 @@ func (syncer *AgentSyncer) checkConfigUpdate() bool {
 			return false
 		}
 		syncer.configRevision = newRevision
+
+		syncer.numaLogger.SetLevel(syncer.config.LogLevel)
```
xdevxy, line 104: Why do we need to set log level here again as we already set it in main?

internal/sync/syncer.go
```diff
@@ -214,6 +214,9 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
 	if err != nil {
 		numaLogger.Error(err, "error getting  the  global config")
 	}
+
+	numaLogger.SetLevel(globalConfig.LogLevel)
```
xdevxy, line 218: same here we can probably not do it?

internal/agent/agent_syncer.go
afugazzotto: I thought to update the log level after every GetConfig call since the level could be changed in the configuration and we want to apply the new level.

xdevxy (approved):

PR #185, opened 2024-03-31
chore: Setup the cluster cache to only watch required resources

### LINKED ISSUE #183: Only cache required resources
# Summary

Currently, we are watching all kinds of resources for cluster cache. To avoid unnecessarily increase the k8s server load and the cache size, we should only cache numaflow resources. 

# Use Cases

Minimize the cache size.

---

xdevxy:
Fixes #183 

### Modifications

This PR changes the cache to only watch Numaflow resources to minimize the scope of the cache. The included resources are configurable via ConfigMap.

### Verification

Verified in a local cluster with the following,
```
make start
```

### DIFF
--- a/internal/sync/cache.go
+++ b/internal/sync/cache.go
@@ -166,7 +166,7 @@ func (c *liveStateCache) getCluster() clustercache.ClusterCache {
 		}
 
 		// TODO: when switch to change event triggering for sync tasking, notify
-		//       the GitSync managing the includedResources that there are changes happen
+		//       the GitSync managing the resources that there are changes happen
 	})
 
 	c.cluster = clusterCache
--- a/internal/sync/cache.go
+++ b/internal/sync/cache.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/client-go/rest"
 
 	"github.com/numaproj-labs/numaplane/internal/common"
+	controllerConfig "github.com/numaproj-labs/numaplane/internal/controller/config"
 	"github.com/numaproj-labs/numaplane/internal/util/kubernetes"
 	"github.com/numaproj-labs/numaplane/internal/util/logger"
 	"github.com/numaproj-labs/numaplane/pkg/apis/numaplane/v1alpha1"
@@ -78,7 +79,7 @@ type LiveStateCache interface {
 	// GetManagedLiveObjs returns state of live nodes which correspond to target nodes of specified gitsync.
 	GetManagedLiveObjs(gitsync *v1alpha1.GitSync, targetObjs []*unstructured.Unstructured) (map[kube.ResourceKey]*unstructured.Unstructured, error)
 	// Init must be executed before cache can be used
-	Init()
+	Init() error
 	// PopulateResourceInfo is called by the cache to update ResourceInfo struct for a managed resource
 	PopulateResourceInfo(un *unstructured.Unstructured, isRoot bool) (interface{}, bool)
 }
@@ -165,7 +166,7 @@ func (c *liveStateCache) getCluster() clustercache.ClusterCache {
 		}
 
 		// TODO: when switch to change event triggering for sync tasking, notify
-		//       the GitSync managing the resources that there are changes happen
+		//       the GitSync managing the includedResources that there are changes happen
 	})
 
 	c.cluster = clusterCache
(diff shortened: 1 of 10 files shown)


internal/sync/cache.go
```diff
@@ -70,15 +71,15 @@ type ResourceInfo struct {
 }
 
 // LiveStateCache is a cluster caching that stores resource references and ownership
-// references. It also stores custom metadata for resources managed by GitSyncs. It always
-// ensures the cache is up-to-date before returning the resources.
+// references. It also stores custom metadata for includedResources managed by GitSyncs. It always
+// ensures the cache is up-to-date before returning the includedResources.
```
juliev0, line 75:
```suggestion
// references. It also stores custom metadata for included resources managed by GitSyncs. It always
// ensures the cache is up-to-date before returning the included resources.
```

juliev0, line 75: was this changed on purpose or by accident?

internal/sync/cache.go
```diff
@@ -322,19 +323,59 @@ func (c *liveStateCache) getSyncedCluster() (clustercache.ClusterCache, error) {
...
 		ResourcesFilter: resourcesFilter,
 	}
 
-	return &cacheSettings{clusterSettings, ignoreResourceUpdatesEnabled}
+	return &cacheSettings{clusterSettings, ignoreResourceUpdatesEnabled}, nil
+}
+
+// parseResourceFilter parse the given rules to generate the
+// included resource to be watched during caching.
+func parseResourceFilter(rules string) (*[]ResourceType, error) {
```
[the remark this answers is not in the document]
juliev0: Also, do you feel confident enough about all of this logic to not need a unit test? (If you feel that your manual test covered all the bases, I'm okay with that)

internal/sync/cache.go
```diff
@@ -363,11 +404,26 @@ func NewNoopNormalizer() diff.Normalizer {
 	return &NoopNormalizer{}
 }
 
-// NoopResourceFilter a noop resource filter
-// TODO: add more meaningful resource filter
-type NoopResourceFilter struct {
+// ResourceFilter resource filter based on what are the includedResources included
```
juliev0, line 407:
```suggestion
// ResourceFilter filter resources based on allowed Resource Types
```

internal/sync/cache.go
xdevxy: yeah this is changed by IDE automatically, will revert it.

internal/sync/cache.go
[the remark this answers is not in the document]
xdevxy: I was actually thinking about a unit test, will add it.

internal/sync/cache.go
```diff
@@ -322,19 +323,59 @@ func (c *liveStateCache) getSyncedCluster() (clustercache.ClusterCache, error) {
...
+	}
 	c.cacheSettings = *setting
+	return nil
 }
 
-func (c *liveStateCache) loadCacheSettings() *cacheSettings {
+func (c *liveStateCache) loadCacheSettings() (*cacheSettings, error) {
 	ignoreResourceUpdatesEnabled := false
-	resourcesFilter := &NoopResourceFilter{}
+	globalConfig, err := controllerConfig.GetConfigManagerInstance().GetConfig()
```
juliev0, line 337: I guess this is not a field that we will allow to dynamically change without needing to restart the Controller?

juliev0: (I'm okay if we don't deal with that here but we may want to have an Issue for that kind of thing, especially if there are other fields like that.)

xdevxy: yeah, right. Dynamic change shouldn't be allowed.

internal/sync/cache.go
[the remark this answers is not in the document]
juliev0: thank you!

internal/sync/cache.go
juliev0: Do you figure there would be issues if we enabled that?

xdevxy: no, I don't see any issue as this config is only read during cache init. So I think we should be fine.

internal/sync/cache.go
```diff
@@ -165,7 +166,7 @@ func (c *liveStateCache) getCluster() clustercache.ClusterCache {
 		}
 
 		// TODO: when switch to change event triggering for sync tasking, notify
-		//       the GitSync managing the resources that there are changes happen
+		//       the GitSync managing the includedResources that there are changes happen
```
juliev0, line 169:
```suggestion
		//       the GitSync managing the resources that there are changes happen
```

internal/sync/cache_test.go
```diff
@@ -199,3 +199,62 @@ metadata:
...
+			resources, err := parseResourceFilter(tc.rules)
+			if tc.hasErr {
+				assert.NotNil(t, err)
+				assert.Nil(t, resources)
+			} else {
+				assert.Nil(t, err)
+				assert.Equal(t, tc.expectedResources, *resources)
+			}
+		})
+	}
```
juliev0, line 259: thanks!

internal/sync/syncer.go
```diff
@@ -112,7 +112,7 @@ func (s *Syncer) StopWatching(key string) {
 }
 
 // Start function starts the synchronizer worker group.
-// Each worker keeps picking up tasks (which contains GitSync keys) to sync the resources.
+// Each worker keeps picking up tasks (which contains GitSync keys) to sync the includedResources.
```
juliev0, line 115:
```suggestion
// Each worker keeps picking up tasks (which contains GitSync keys) to sync the resources.
```

internal/sync/diff.go
```diff
@@ -136,7 +136,7 @@ func NewIgnoreNormalizer(overrides map[string]ResourceOverride) (diff.Normalizer
 	return &ignoreNormalizer{patches: patches}, nil
 }
 
-// Normalize removes fields from supplied resource using json paths from matching items of specified resources ignored differences list
+// Normalize removes fields from supplied resource using json paths from matching items of specified includedResources ignored differences list
```
juliev0, line 139:
```suggestion
// Normalize removes fields from supplied resource using json paths from matching items of specified resources ignored differences list
```

internal/sync/syncer.go
```diff
@@ -267,7 +270,7 @@ func (r *resourceInfoProviderStub) IsNamespaced(_ schema.GroupKind) (bool, error
 	return false, nil
 }
 
-// sync compares the live state of the watched resources to target state defined in git
+// sync compares the live state of the watched includedResources to target state defined in git
```
juliev0, line 273:
```suggestion
// sync compares the live state of the watched resources to target state defined in git
```

xdevxy: Oops sorry for these unintended changes. Thanks for catching it!

juliev0 (approved):

PR #186, opened 2024-03-31
Update Controller ClusterRole

### LINKED ISSUE #184: Limit the power of Controller ClusterRole
# Summary

Currently the controller requires super user for ClusterRole, we should limit it to what's actually required and follow the least privilege principle.

# Use Cases

To avoid expose unnecessary power to Numaplane controller.

---

xdevxy:
Fixes #184 

### Modifications
This PR updates the Controller ClusterRole to no longer requires super user permission, instead only requires whats needed for manage Numaflow resources. 

### Verification

Verified in a local cluster to deploy a Numaflow pipeline.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -4,9 +4,28 @@ kind: ClusterRole
 metadata:
   name: numaplane-role
 rules:
-- apiGroups:
-  - '*'
+- apiGroups: ["numaflow.numaproj.io"]
+  resources: ["*"]
+  verbs: ["*"]
+- apiGroups: ["numaplane.numaproj.io"]
+  resources: ["*"]
+  verbs: ["*"]
+- apiGroups: [""]
   resources:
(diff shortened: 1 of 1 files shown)


config/rbac/role.yaml
```diff
@@ -4,9 +4,28 @@ kind: ClusterRole
...
+  resources:
+    - rolebindings
+    - roles
+  verbs:
+    - '*'
+- apiGroups: ["apps"]
+  resources:
+    - deployments
+  verbs:
+    - '*'
```
juliev0, line 31: Oh, I guess `Deployment` needs to be here because of the Numaflow Controller, huh? Did you verify your list includes all of the types that are part of the Numaflow Controller manifest?

xdevxy: yeah I verified. :)

juliev0 (approved):

Issue #182, opened 2024-03-31
Error on syncing InterStepBufferService


xdevxy:
**Describe the bug**
https://github.com/numaproj-labs/numaplane/pull/180 seems to introduce a syncing error as the following,

[...]

PR #187, opened 2024-04-02
Logger fix

afugazzotto:
Fixes https://github.com/numaproj-labs/numaplane/issues/182

### Modifications

The log level mapping did not work as understood when packages call logr.Logger.Info func directly without calling the NumaLogger.Info or other function. A distinction had to be made between the 2 types of calls. For this reason, the wrapper functions (Info, Debug, etc.) call the sink Info function directly instead of setting the verbosity and then calling the Logger Info func. In addition, levels map was changed to solve additional problems with configmap not having set a log level resulting in a conflicting "zero" level. Other minor adjustments were made to avoid logging the entire gitsync manifest.

### Verification

Additional unit tests were added. Additional manual comparisons between "old logs" (prior to commit https://github.com/numaproj-labs/numaplane/commit/0d0a43dadae9e3dd830c20b62bc149119a9fd6e6) and "new logs" (with current PR changes) were performed.

### DIFF
--- a/internal/util/logger/logger.go
+++ b/internal/util/logger/logger.go
@@ -47,9 +47,9 @@ const (
 
 // The following map define the logr verbosity or NumaLogger semantic levels (constants above) mapping
 // to the zerolog levels (https://github.com/rs/zerolog/blob/master/log.go#L129).
-// This conversion/mapping is due to the following relationship between zerolog levels and logr verbosity:
-// - zerolog(4=Fatal) = "always want to see" => zerolog(-1=Trace) = "only want to see rarely, ex: for debug purposes"
-// - logrVerbosity(0) = "always want to see" => logrVerbosity(10) = "only want to see rarely, ex: for debug purposes"
+// This conversion/mapping is due to the following inverse relationship between zerolog levels and logr verbosity:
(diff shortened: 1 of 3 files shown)


internal/util/logger/logger.go
```diff
@@ -20,28 +20,36 @@ const (
...
+|    warn             |    2            |    2           |
+|    info             |    3            |    1           |
+|    debug            |    4            |    0           |
+|    verbose          |    5            |    -2 (custom) |
 */
-
 const (
-	FatalLevel   = 0
-	WarnLevel    = 1
-	InfoLevel    = 2
```
xdevxy, line 40: If specify `logLevel: 4` in configmap, I should expect Debug level logging, right?

afugazzotto: Yes, that is the expected behavior.

xdevxy: Thanks! I verified that works. However, I also tried to see if log level can be dynamically configured. That is working except one scenario, if the logLevel field is removed from the configmap, we are still setting the [value](https://github.com/numaproj-labs/numaplane/blob/main/internal/sync/syncer.go#L223) instead of using default one. I think this is minor can be address as a follow up. do you mind create an issue to track it? Thanks a lot!

xdevxy (approved):

PR #188, opened 2024-04-02
e2e test suite improvements

dpadhiar:
Add small fixes to e2e framework and adds target in Makefile to run tests. 
Functional test file created with sample repo files, test not fully written.
Pushing to gitserver fixed with changes from @shubhamdixit863 
Include file to port-forward git-server pod
Before/AfterTest functions for cleanup after each test case to ensure repo is clean

### Verification

Ran `make test-e2e` to see that test fully succeeds in sending a push request twice.


Makefile
```diff
@@ -208,3 +208,10 @@ gitserver:
 		--platform $(PLATFORMS) \
 		-t $(GITSERVER_IMAGE) \
 		--push .
+
+test-e2e:
+test-%:
+	kubectl -n numaplane-system delete pod localgitserver-0 --ignore-not-found=true
```
juliev0, line 214: I believe it's an entire StatefulSet that needs to be deleted, right? not just the Pod?

juliev0: and the `Service` as well.
Perhaps, since below you run `kubectl apply -n numaplane-system -k ./tests/e2e-gitserver`, does it work to first do `kubectl delete -n numaplane-system -k ./tests/e2e-gitserver`? That might be nice and clean.

dpadhiar: Just tried this, works just fine. Will make change.

tests/fixtures/e2e_suite.go
```diff
@@ -82,6 +84,61 @@ func (s *E2ESuite) TearDownSuite() {
...
+	}
+
+	// git push to remote
+	err = repo.Push(&git.PushOptions{
+		RemoteName: "origin",
+		Auth:       auth,
+	})
+	if err != nil {
+		s.CheckError(err)
+	}
```
juliev0, line 135: nit: can you break this out into a function called `ResetRepository()` or something?

juliev0:
So, is this correct?:
When the Suite of tests starts up, you [clone the repository and add in all of the test files?](https://github.com/numaproj-labs/numaplane/pull/188/files#diff-fad3bf7ace79903d82772107a508fe0a1478ce3eac71d9bde39aed59a1c5c5e8R35)

But then after each test you clear the remote repository of all of those files?

dpadhiar:
> So, is this correct?: When the Suite of tests starts up, you [clone the repository and add in all of the test files?](https://github.com/numaproj-labs/numaplane/pull/188/files#diff-fad3bf7ace79903d82772107a508fe0a1478ce3eac71d9bde39aed59a1c5c5e8R35)
> 
> But then after each test you clear the remote repository of all of those files?

When the suite starts, the Git server is started and (as currently set) 5 repos are initialized on it (repo1.git etc). 

For each test, we create the Given struct with the provided GitSync. This GitSync's Path (sample-pipeline) is used to clone the repo (localhost:8080/git/repo1.git) into local. We copy the test files based on the path as well so the end result for local is `local/sample-pipeline`. We push those to the Git server.

When has a function to then create the GitSync which will read from the Git server. It can also makes its own commits by providing the directory (ex: sample-pl-commits) and file.

AfterTest will clean up the local repo as well as the Git server to ensure the next test case has a clean working repo.

juliev0:
> > So, is this correct?: When the Suite of tests starts up, you [clone the repository and add in all of the test files?](https://github.com/numaproj-labs/numaplane/pull/188/files#diff-fad3bf7ace79903d82772107a508fe0a1478ce3eac71d9bde39aed59a1c5c5e8R35)
> > But then after each test you clear the remote repository of all of those files?
> 
> When the suite starts, the Git server is started and (as currently set) 5 repos are initialized on it (repo1.git etc).
> 
> For each test, we create the Given struct with the provided GitSync. This GitSync's Path (sample-pipeline) is used to clone the repo (localhost:8080/git/repo1.git) into local. We copy the test files based on the path as well so the end result for local is `local/sample-pipeline`. We push those to the Git server.
> 
> When has a function to then create the GitSync which will read from the Git server. It can also makes its own commits by providing the directory (ex: sample-pl-commits) and file.
> 
> AfterTest will clean up the local repo as well as the Git server to ensure the next test case has a clean working repo.

Thanks for explaining. Can you as part of this change include your overall plan for where stuff goes, like when you expand to include other tests? Might be helpful to see a high level overview, unless that's already in the Design Document?

dpadhiar:
> > > So, is this correct?: When the Suite of tests starts up, you [clone the repository and add in all of the test files?](https://github.com/numaproj-labs/numaplane/pull/188/files#diff-fad3bf7ace79903d82772107a508fe0a1478ce3eac71d9bde39aed59a1c5c5e8R35)
> > > But then after each test you clear the remote repository of all of those files?
> > 
> > 
> > When the suite starts, the Git server is started and (as currently set) 5 repos are initialized on it (repo1.git etc).
> > For each test, we create the Given struct with the provided GitSync. This GitSync's Path (sample-pipeline) is used to clone the repo (localhost:8080/git/repo1.git) into local. We copy the test files based on the path as well so the end result for local is `local/sample-pipeline`. We push those to the Git server.
> > When has a function to then create the GitSync which will read from the Git server. It can also makes its own commits by providing the directory (ex: sample-pl-commits) and file.
> > AfterTest will clean up the local repo as well as the Git server to ensure the next test case has a clean working repo.
> 
> Thanks for explaining. Can you as part of this change include your overall plan for where stuff goes, like when you expand to include other tests? Might be helpful to see a high level overview, unless that's already in the Design Document?

I can include something new in the Design Doc to show how that might work. For now, it should be possible to create multiples different paths to different Numaflow files in `local` using the `CloneGitRepo` function. This requires changing the path in the GitSync currently.

tests/fixtures/given.go
```diff
@@ -110,29 +113,49 @@ func (g *Given) readResource(text string, v metav1.Object) {
 
 // initialize git repo with all Numaflow files
 func (g *Given) CloneGitRepo() *Given {
```
juliev0, line 115: maybe we should rename this function to something like `InitializeGitRepo()` since it does more than just clone, it actually pushes files too

juliev0: also, if in particular it's initializing with particular files, a comment would be helpful to indicate that

juliev0 (approved):

xdevxy (approved):

PR #193, opened 2024-04-04
chore: dynamically pick up config change for cache setting

### LINKED ISSUE #191: Cache Setting should dynamically pick up any config changes
# Summary

What change needs making?

Currently, we only load the cluster cache setting during cache init. However, from users perspective, to avoid confusion, any changes to configmap should be all be dynamically picked up or requires restart. To be coherent with other configs, we should also 
pick up any config changes for cache setting dynamically as well.

When would you use this?
Update cache setting configs, specifically the `includedResources` rules.

---

xdevxy:
Fixes #191 

### Modifications

This PR updates it to dynamically pick up config change for cache setting.

### Verification

Verified in a local cluster by update the `includedResources` field in config map and see the cache setting is reloaded.

### DIFF
--- a/internal/sync/cache.go
+++ b/internal/sync/cache.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/url"
 	"os/exec"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -17,7 +18,6 @@ import (
 	"github.com/argoproj/gitops-engine/pkg/health"
 	"github.com/argoproj/gitops-engine/pkg/utils/kube"
 	"github.com/cespare/xxhash/v2"
-	"go.uber.org/zap"
 	"golang.org/x/sync/semaphore"
 	v1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


juliev0:
can you look at the unit test error?

xdevxy:
> can you look at the unit test error?

fixed, thanks!

internal/sync/cache.go
```diff
@@ -120,7 +122,36 @@ func NewLiveStateCache(
 	}
 }
 
+func (c *liveStateCache) invalidate(cacheSettings cacheSettings) {
+	c.logger.Info("invalidating live state cache")
+	c.lock.Lock()
+	c.cacheSettings = cacheSettings
+	cluster := c.cluster
```
juliev0, line 129: is this variable not used?

xdevxy: you are asking about `cluster`? it is used at L132 .

juliev0: oh, I see - the way the indentation is showing up in my browser on line 132 is weird because it's trying to squeeze the whole line in, and it looks like the start of a new function.

juliev0 (approved):

Issue #190, opened 2024-04-04
e2e test: syncer cannot clone test git server repos


dpadhiar:
**Describe the bug**

The `localgitserver` pod is used in e2e tests to create Git repos that can be pushed and pulled to for properly testing Numaplane operations. However if you create a GitSync with a repoURL for one of the repos (ie. http://localhost:8080/git/repo1.git) while the pod is port forwarded, the resources will not be created in the target namespace and the controller logs will show this error:
```
{"level":"error","ts":"2024-04-03T16:41:16.254379412Z","logger":"numaplane.synchronizer","caller":"sync/syncer.go:183","msg":"Failed to execute a task","gitSyncKey":"numaplane-system/gitsync-example","error":"failed to clone the repo of key \"numaplane-system/gitsync-example\", repository not found","stacktrace":"github.com/numaproj-labs/numaplane/internal/sync.(*Syncer).run\n\t/workspace/internal/sync/syncer.go:183"}
```

The `repository not found` error may be occurring due to auth permissions not being correct. It is still possible to `git clone` the repo on the Terminal, verifying that the repo does exist and is reachable.

**To Reproduce**

1.  Manually create and port-forward localgitserver pod
2. Create GitSync with `repoURL: http://localhost:8080/git/repo1.git`
3. Check controller logs

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

PR #194, opened 2024-04-05
Fixed e2e tests configurations

shubhamdixit863:
Yes it fixes a bug

Fixes  https://github.com/numaproj-labs/numaplane/issues/190

### Modifications

**Configuration Manifests**: Introduced separate configuration manifests folder for  numaplane controller  for running e2e tests

**Makefile Enhancements**:
`e2e-test-clean`: Cleans up resources after tests to ensure a clean environment for subsequent test runs.
`e2e-test-start`: Prepares the environment for e2e testing, including setting up the NumaPlane Controller.

**GitSync YAML Update**: Modified to use Fully Qualified Domain Names (FQDN) for git urls.
**Kubernetes Secrets**: Added Kubernetes secrets to the GitSync Controller's configuration, to be used by the internal gitsync to sync repositories

### Verification

These improvements were thoroughly tested in a local cluster using the `make test-e2e` command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,9 +209,21 @@ gitserver:
 		-t $(GITSERVER_IMAGE) \
 		--push .
 
+.PHONY: e2e-test-clean
+e2e-test-clean:
+	$(KUBECTL) delete -k tests/e2e/manifests --ignore-not-found=true
+	$(KUBECTL) delete -f tests/e2e/manifests/numaplane-ns.yaml --ignore-not-found=true
+
+.PHONY: e2e-test-start
+e2e-test-start: e2e-test-clean image
+	$(KUBECTL) apply -f tests/e2e/manifests/numaplane-ns.yaml
+	$(KUBECTL) kustomize tests/e2e/manifests | sed 's/CLUSTER_NAME_VALUE/$(CLUSTER_NAME)/g' | sed 's@quay.io/numaproj/@$(IMAGE_NAMESPACE)/@' | sed 's/$(IMG):$(BASE_VERSION)/$(IMG):$(VERSION)/' | $(KUBECTL) apply -f -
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -209,9 +209,21 @@ gitserver:
...
 		--push .
 
+.PHONY: e2e-test-clean
+e2e-test-clean:
+	$(KUBECTL) delete -k tests/e2e/manifests --ignore-not-found=true
+	$(KUBECTL) delete -f tests/e2e/manifests/numaplane-ns.yaml --ignore-not-found=true
+
+.PHONY: e2e-test-start
+e2e-test-start: e2e-test-clean image
+	$(KUBECTL) apply -f tests/e2e/manifests/numaplane-ns.yaml
```
dpadhiar, line 219: We should also add a step to delete and create `numaflow-pipeline` namespace as that is the current target namespace for GitSync e2e testing. We could consider making this variable in the future.

tests/fixtures/given.go
```diff
@@ -38,6 +38,10 @@ import (
 	planepkg "github.com/numaproj-labs/numaplane/pkg/client/clientset/versioned/typed/numaplane/v1alpha1"
 )
 
+// localGitUrl is set for local development/testing,
+// the GitSync controller uses a different URL configured in GitSync yaml.
+const localGitUrl = "http://localhost:8080/git/repo1.git"
```
dpadhiar, line 43: We could make this line `http://localhost:8080/git/%s` and trim `http://localgitserver-service.numaplane-system.svc.cluster.local` from the RepoURL to create the right URL for each repo (ie. repo2.git if we use it)

juliev0 (approved):

PR #204, opened 2024-04-11
Added valid yaml file to docker repo

shubhamdixit863:
No

### Modifications

Modified the Dockerfile such that git repos should have  proper kubernetes resource.yaml files. As currently gitsync throws error  while syncing the repos as the resource yaml is invalid

### Verification

Tested in local kubernetes cluster


juliev0:
@shubhamdixit863 @dpadhiar Is it necessary for the Dockerfile to commit any files at all? Might be nice if this repository is a little more generic. Let me know if it creates difficulties not to. I know it does mean in that case that you need a client to push to it. Does it create difficulties?

juliev0:
Also, want to clarify what the state you saw was @dpadhiar - was it that the GitSync basically showed an Error in the Status?

dpadhiar:
> Also, want to clarify what the state you saw was @dpadhiar - was it that the GitSync basically showed an Error in the Status?

I'm unsure what error this is preventing at this time. This may be the case where there is no path and the GitSync tried applying `data.yaml` which isn't a valid object. In regards to not pushing a file at all while creating a repo, @shubhamdixit863 will know more if it is entirely necessary.

shubhamdixit863:
> > Also, want to clarify what the state you saw was @dpadhiar - was it that the GitSync basically showed an Error in the Status?
> 
> I'm unsure what error this is preventing at this time. This may be the case where there is no path and the GitSync tried applying `data.yaml` which isn't a valid object. In regards to not pushing a file at all while creating a repo, @shubhamdixit863 will know more if it is entirely necessary.

@juliev0 yes thats right data.yaml was not a valid file by default so when syncing it gives error ,its not necessary but just for convinience . Moreover its generic you can create repos and push and pull into and from them . I guess @dpadhiar you have already tested it .

shubhamdixit863:
Closing as raised another PR -https://github.com/numaproj-labs/numaplane/pull/208

PR #205, opened 2024-04-11
functional e2e test case

### LINKED ISSUE #171: write functional e2e test for GitSync using raw manifests

dpadhiar:
Fixes #171 

### Modifications

Renamed targetNamespace to `numaplane-e2e` as `numaflow-pipeline` would not be accurate if we were to test using other resources.

Makefile changes to allow user to run tests using `make test-e2e` with proper cleanup and setup of necessary namespaces and components including Numaplane controller and git server pod.

Test config manifests for e2e testing have been updated for "root" user which is the correct credential needed to access Git server pod repos.

Created two new directories: `testdata/numaflow` and `testdata/basic-resources` - These include an initial commit dir with all files needed to initialize the repo and modified dir which contains all changes we'd like to push to the Git repo. This can be further nested. Future tests can be expected to follow this structure to host test manifests.

Functional test case for CRUD operations for Numaflow resources and basic k8s resources.
Self healing test case for basic k8s resources.
ChangeRepoURL test to change both repoUrl or path of a GitSync.

Added / updated functions to:
-  verify that the spec of a resource has been changed or not. This requires getting the resource using kubeClient and supplying the key:value to check.
- verify that the commitStatus of the GitSync is `synced: true` and that the Hash is up to date with the most recent commit created by the test. Each struct stores this commit hash now.
- update a GitSync as opposed to creating a new one, this will be needed for testing changing the repoPath of a GitSync
- `PushToGitRepo` now allows user to delete files from repo as well. This is necessary for testing that GitSync will delete resources when synced with repo

TODO: helm manifests

### Verification

Ran `make test-e2e` to see tests are passing.
Port-forwarded Git server pod and created GitSync manually to note that expected actions were occurring.


Makefile
```diff
@@ -211,18 +211,22 @@ gitserver:
 
 .PHONY: e2e-test-clean
 e2e-test-clean:
-	$(KUBECTL) delete -k tests/e2e/manifests --ignore-not-found=true
-	$(KUBECTL) delete -f tests/e2e/manifests/numaplane-ns.yaml --ignore-not-found=true
+	$(KUBECTL) delete -n numaplane-e2e isbsvc --all
+	$(KUBECTL) delete -n numaplane-e2e pipelines --all
+	$(KUBECTL) delete -n numaplane-e2e cm --all
+	$(KUBECTL) delete -n numaplane-e2e all --all
```
juliev0, line 217: I guess you're no longer including these resources. What happens if you just have `$(KUBECTL) delete -n numaplane-e2e all --all` instead of the 3 lines above?

tests/e2e/functional_test.go
```diff
@@ -30,12 +31,162 @@ type FunctionalSuite struct {
...
+		When().
+		CreateGitSyncAndWait()
+	defer w.DeleteGitSyncAndWait()
+
+	w.Expect().ResourcesExist("apps/v1", "deployments", []string{"test-deploy"})
+	w.Expect().ResourcesExist("v1", "configmaps", []string{"test-config"})
+	w.Expect().ResourcesExist("v1", "secrets", []string{"test-secret"})
+	w.Expect().CheckCommitStatus()
+
+	// w.PushToGitRepo("kustomize/initial-commit", []string{"deployment.yaml"}, true)
```
dpadhiar, line 184: This is commented out as I was not seeing the deployment getting deleted when syncing with the commit.

Makefile
juliev0: oh wait, I think I'm wrong about that...looks like you are using numaflow resources...sorry

juliev0: That's weird about the ConfigMaps. Did you find any corroboration of that on the internet by chance?

> I was unsure about the whole installing the CRDs question for the CI

But you're still going to need it in the CI if it's in the one test case, right?

juliev0: Also, why do the isbsvc's and pipelines need to be deleted first? Is it because you're installing a Numaflow Controller, and those need to have their finalizer removed before the Controller is deleted?

tests/fixtures/expect.go
```diff
@@ -34,38 +37,117 @@ type Expect struct {
...
+			Resource(resourceType).
+			Name(resource).
+			DoRaw(ctx)
+	} else {
+		body, err = e.kubeClient.CoreV1().RESTClient().Get().AbsPath(
+			fmt.Sprintf("/apis/%s/namespaces/%s/%s/%s",
+				apiVersion,
+				TargetNamespace,
+				resourceType,
+				resource)).DoRaw(ctx)
```
juliev0, line 105: Just curious - does this call not work for CoreV1 Resources?

dpadhiar: Yes, I wasn't able to `Get` the example configMaps or secrets I was creating.

tests/fixtures/expect.go
```diff
@@ -34,38 +37,117 @@ type Expect struct {
...
+			e.t.Fatalf("Resource %s/%s not deleted", resourceType, r)
+		}
+		e.t.Logf("Resource %s/%s doesn't exist", resourceType, r)
+	}
+
+	return e
+}
+
+// verify value of resource spec to determine if change occurred or not as expected
+func (e *Expect) VerifyResourceState(apiVersion, resourceType, resource, field, key string, value interface{}) *Expect {
```
juliev0, line 76: want to clarify in a comment that it currently only works for top-level fields? (I think?)

tests/fixtures/expect.go
```diff
@@ -34,38 +37,117 @@ type Expect struct {
...
+// verify that GitSync's commitStatus is as expected:
+// synced = true
+// hash is equal to the last commit to the Git server
+func (e *Expect) CheckCommitStatus() *Expect {
+
+	e.t.Helper()
+	e.t.Log("Verifying GitSync's commitStatus is as expected..")
+
+	// gitSync commit status will be nil if reconciliation has not finished
+	commitStatus, err := e.getCommitStatus()
```
juliev0, line 140: I see there's a timeout in `getCommitStatus()` to allow some time for the `CommitStatus != nil`. I'm kind of wondering if the timeout should be over here, though. Consider that initially a GitSync will have no CommitStatus, then after the first reconciliation, there's a CommitStatus. Then something changes in the synced resources, and the CommitStatus gets updated to the new one. If we call `CheckCommitStatus()` after that last time, it will immediately return from `getCommitStatus()` because there is a CommitStatus (the old one maybe), and then here it may find that the hash doesn't match. Let me know if I have that wrong.

Makefile
juliev0: Good find! That's crazy, though. Can you add comments to indicate why you're deleting the way you are in case someone's wondering later? Maybe also link to the comment wrt the ConfigMap?

tests/fixtures/expect.go
juliev0: okay, thanks

tests/fixtures/given.go
```diff
@@ -40,14 +40,15 @@ import (
 
 // localGitUrl is set for local development/testing,
 // the GitSync controller uses a different URL configured in GitSync yaml.
-const localGitUrl = "http://localhost:8080/git/repo1.git"
+// const localGitUrl = "http://localhost:8080/git/repo1.git"
```
juliev0, line 43:
```suggestion
```

juliev0, line 43: delete?

tests/fixtures/given.go
```diff
@@ -40,14 +40,15 @@ import (
...
 
 var (
 	auth = &http.BasicAuth{
 		Username: "root",
 		Password: "root",
 	}
-	localPath = "./local"
+	localPath   = "./local"
+	localGitUrl = "http://localhost:8080/git/%s"
```
juliev0, line 51: add your comments here instead?

Makefile
dpadhiar: Will do!

tests/fixtures/expect.go
dpadhiar: This is a good callout, I need to revisit this method tomorrow. Initially when I was checking for `CommitStatus` I was unsure why I kept getting a nil pointer and thats why I included the timeout there at the time. Would timing out on the condition that the `hash` is as we expected better? This assumes `Synced == true` as well.

tests/fixtures/given.go
dpadhiar: Will do! Missed on cleanup.

tests/fixtures/given.go
```diff
@@ -142,30 +147,46 @@ func (g *Given) InitializeGitRepo() *Given {
...
+				err := CopyFile(filepath.Join(dataPath, name, dirFile.Name()), filepath.Join(tmpPath, name, dirFile.Name()))
+				if err != nil {
+					g.t.Fatal(err)
+				}
+			}
+		} else {
+			err := CopyFile(filepath.Join(dataPath, name), filepath.Join(tmpPath, name))
+			if err != nil {
+				g.t.Fatal(err)
+			}
```
juliev0, line 179: Is the idea really to recursively copy the entire directory over? Because [os.Copy() ](https://pkg.go.dev/github.com/otiai10/copy#section-readme)apparently does that.

dpadhiar: Looks like this is a different package but this would be exactly what we need to copy over that initial commit into the local repo folder (and replace the function I had to pull to CopyFiles as is). I had made these changes to this function for the case that we needed to test a kustomize repo which would more than likely have multiple subdirectories. The same goes for Helm so this would help a ton.

tests/fixtures/expect.go
dpadhiar: Sure thing, I'll come up with a definitive answer tomorrow.

tests/e2e/functional_test.go
```diff
@@ -30,12 +31,162 @@ type FunctionalSuite struct {
...
+// GitSync testing for Numaflow with raw manifests
+func (s *FunctionalSuite) TestSimpleNumaflowGitSync() {
+
+	// create GitSync and initialize Git repo with initial manifest files
+	w := s.Given().GitSync("@testdata/gitsync.yaml").InitializeGitRepo("numaflow/initial-commit").
 		When().
 		CreateGitSyncAndWait()
 	defer w.DeleteGitSyncAndWait()
+
+	// verify Numaflow installation and pipeline creation
```
juliev0, line 43: is the "Numaflow installation" part of the comment outdated now?

tests/e2e/functional_test.go
```diff
@@ -30,12 +31,162 @@ type FunctionalSuite struct {
...
+	w.Expect().ResourcesExist("numaflow.numaproj.io/v1alpha1", "pipelines", []string{"simple-pipeline"})
+	// w.Expect().VerifyResourceSpec("numaflow.numaproj.io/v1alpha1", "pipelines", "simple-pipeline", "vertices", "")
+	w.Expect().CheckCommitStatus()
+
+	// edit existing pipeline back to initial state
+	w.PushToGitRepo("numaflow/initial-commit", []string{"sample_pipeline.yaml"}, false).Wait(30 * time.Second)
+
+	w.Expect().ResourcesExist("numaflow.numaproj.io/v1alpha1", "pipelines", []string{"simple-pipeline"})
+	w.Expect().ResourcesDontExist("numaflow.numaproj.io/v1alpha1", "vertices", []string{"simple-pipeline-another-out"})
+	w.Expect().ResourcesExist("numaflow.numaproj.io/v1alpha1", "vertices", []string{"simple-pipeline-in", "simple-pipeline-cat", "simple-pipeline-out"})
```
juliev0, line 71: I don't think you can do these tests involving vertices, right? this requires a Numaflow Controller to be running

tests/e2e/functional_test.go
dpadhiar: I suppose technically not since we're deploying the controller still. I will change the comment though since that isn't what we're looking to test for, just that the manifests are being applied.

tests/e2e/functional_test.go
dpadhiar: That's right, these can be removed in favor of checking the spec.

juliev0:
I didn't look closely...did you verify all of these?:

- deleted resources should include both deleted files, and files which include multiple resources in which a subset of the resources were deleted
- added resources should include both new files, and files which include multiple resources in which new resources were added to it

dpadhiar:
> I didn't look closely...did you verify all of these?:
> 
> * deleted resources should include both deleted files, and files which include multiple resources in which a subset of the resources were deleted
> * added resources should include both new files, and files which include multiple resources in which new resources were added to it

Just missing deleting a subset of a file defining multiple resources, will add. If you look at the `BasicGitSync` test again, there is a file called `multiple-resources` defining a deployment and configmap.

tests/e2e/functional_test.go
```diff
@@ -30,12 +31,162 @@ type FunctionalSuite struct {
...
+	w.Expect().CheckCommitStatus()
+
+	// old resources are deleted as their manifests do not exist in new repository
+	w.Expect().ResourcesDontExist("apps/v1", "deployments", []string{"test-deploy"})
+	w.Expect().ResourcesDontExist("v1", "configmaps", []string{"test-config"})
+	w.Expect().ResourcesDontExist("v1", "secrets", []string{"test-secret"})
+
+}
+
+func (s *FunctionalSuite) TestKustomize() {
```
dpadhiar, line 172: @juliev0 Technical question for kustomize, is the GitSync watching for the resource.yaml file to be deleted or the resource being deleted from the kustomization.yaml files? (ie. kustomize file at the base dir and kustomize file where the resource files are)

juliev0: Let's talk about what _should_ be rather than what _is_ since that's what matters. Basically, Numaplane should run kustomize on the kustomization.yaml file, so whatever resources are named in there are what matters. Does it answer your question or no?

But one clarification - if the kustomization.yaml file references another repository and something changes in that repository, we won't be notified of the change or anything. So, we won't react to it. That was a design decision made I guess. But if something in our own repo changes, then we'll always run kustomize on whatever is latest and greatest, so that's when we'd notice a change made within that repo.

dpadhiar: That does answer my question then, I need to look into why the deployment for example isn't being deleted when it's removed from the kustomization file in that case.

Will the GitSync just run the kustomize file again when it sees any changes to the manifests on the repo?

juliev0: Good, maybe you've found another bug. Want to explore the code a little from here?: https://github.com/numaproj-labs/numaplane/blob/main/internal/sync/syncer.go#L232 You should find some code related to kustomize manifests in there

dpadhiar: Will do! If this is in fact a bug, should we look to get that fixed first before pushing this or getting this in sooner? The test could be commented out for now and be included in the followup PRs that I had discussed some with Hao.

tests/fixtures/given.go
dpadhiar: Replaced.

tests/e2e/functional_test.go
juliev0: I'm okay either way. Just make sure a bug gets logged first

dpadhiar: I'm not positive just yet, I think this may be something Chandan is fixing here: https://github.com/numaproj-labs/numaplane/pull/216

Going to ask if he can test his PR with kustomize repo.

xdevxy: @dpadhiar why do you need to do L184? Are you trying to remove `deployment.yaml`?

@chandankumar4 I believed you test if kustomize file changed the resources should be added/deleted/updated accordingly? And we are checking each time with kustomize build the end yaml file right? So even there is a remote referred repository we can detect the change, right?

dpadhiar: @xdevxy L184 is leftover from me testing if the resource wasn't being deleted because the manifest was still on the repo. That can be removed once @chandankumar4 change is merged.

Makefile
```diff
@@ -209,20 +209,28 @@ gitserver:
 		-t $(GITSERVER_IMAGE) \
 		--push .
 
+# Pipelines and ISBs are deleted first as they may have finalizers which will cause delete all to hang
```
juliev0, line 212:
```suggestion
# Pipelines and ISBs are deleted first as they may have finalizers which will cause "delete all" to hang (because if a Controller is also being deleted, it first needs to remove the finalizers before it's deleted)
```

Makefile
```diff
@@ -209,20 +209,28 @@ gitserver:
...
+# ConfigMaps are explicitly deleted as they are not included in `kubectl delete all`
```
juliev0, line 213:
```suggestion
# ConfigMaps and Secrets are explicitly deleted as they are not included in `kubectl delete all`
```

tests/fixtures/e2e_suite.go
```diff
@@ -185,25 +180,56 @@ func resetRepo() error {
...
+			if err != nil {
+				return err
+			}
+		}
 	}
+
 	return nil
 }
+
+// func (s *E2ESuite) resetNamespace() error {
```
juliev0, line 232: is this something to be removed or done later?

dpadhiar: Removed.

tests/fixtures/expect.go
```diff
@@ -34,52 +37,163 @@ type Expect struct {
...
+		if !e.doesNotExist(apiVersion, resourceType, r) {
+			e.t.Fatalf("Resource %s/%s not deleted", resourceType, r)
 		}
+		e.t.Logf("Resource %s/%s doesn't exist", resourceType, r)
 	}
 
 	return e
 }
 
+// verify value of resource spec to determine if change occurred or not as expected
```
juliev0, line 75:
```suggestion
// verify value of resource spec to determine if field is set
```

juliev0: (because you could also call this to verify something didn't change)

tests/fixtures/expect.go
```diff
@@ -34,52 +37,163 @@ type Expect struct {
...
+		e.t.Fatalf("Failed to verify resource %s/%s state", resourceType, resource)
+	}
+
+	for k, v := range object[field].(map[interface{}]interface{}) {
+		if k == key {
+			if v != value {
+				e.t.Fatalf("Resource %s/%s state is not as expected", resourceType, resource)
+			}
+		}
+	}
```
juliev0, line 124: I wonder how hard it would be to make this a little better. Right now it's basically verifying "field.key", which assumes that the value to verify is exactly 2 levels down. How hard would it be to change it so that it's just a "key" (no "field") and its value is demarcated by periods? Then you'd need to have a recursive call here.

dpadhiar: `w.Expect().VerifyResourceState("apps/v1", "deployments", "test-deploy", "spec.replicas", 3)`

In that case, for every "." we would just recurse further until we get to the value we want to check.

juliev0: The one complexity is that you may have someone who wants to check arrays as well like `spec.myArray[1].myKey`, so ideally it would handle arrays as well. I wonder if there's some gomodule out there which can do something like this.

dpadhiar: That would be helpful, something that can help iterate over a `map[string]interface{}` with many more `interface{}`

juliev0:
> > I didn't look closely...did you verify all of these?:
> > 
> > * deleted resources should include both deleted files, and files which include multiple resources in which a subset of the resources were deleted
> > * added resources should include both new files, and files which include multiple resources in which new resources were added to it
> 
> Just missing deleting a subset of a file defining multiple resources, will add. If you look at the `BasicGitSync` test again, there is a file called `multiple-resources` defining a deployment and configmap.

I didn't check...have you added it now?

dpadhiar:
> > > I didn't look closely...did you verify all of these?:
> > > 
> > > * deleted resources should include both deleted files, and files which include multiple resources in which a subset of the resources were deleted
> > > * added resources should include both new files, and files which include multiple resources in which new resources were added to it
> > 
> > 
> > Just missing deleting a subset of a file defining multiple resources, will add. If you look at the `BasicGitSync` test again, there is a file called `multiple-resources` defining a deployment and configmap.
> 
> I didn't check...have you added it now?

Added now.

tests/fixtures/expect.go
dpadhiar: Thanks! Looking into this

dpadhiar: Doesn't look like it's available as a package, I might be able to mock what they're doing to parse the YAML though.

xdevxy (approved):

tests/fixtures/expect.go
```diff
@@ -34,52 +38,140 @@ type Expect struct {
...
+			e.t.Fatalf("Resource %s/%s not deleted", resourceType, r)
 		}
+		e.t.Logf("Resource %s/%s doesn't exist", resourceType, r)
 	}
 
 	return e
 }
 
+// verify value of resource spec to determine if field is set
+func (e *Expect) VerifyResourceState(apiVersion, resourceType, resource, field, key string, value interface{}) *Expect {
```
juliev0, line 77:
```suggestion
// TODO: this method currently only handles fields that are 2 levels down from root 
// (i.e. "field.key"), and should be modified to support keys at any level, like:
// - "spec.metadata.labels.instance"
// - "spec.template.containers[0].name"
func (e *Expect) VerifyResourceState(apiVersion, resourceType, resource, field, key string, value interface{}) *Expect {
```

PR #215, opened 2024-04-17
chore: further reduce the required permission

xdevxy:
### Modifications

This PR did a few improvements/clean up:

-  Further reduced the required permission in ClusterRole
-  Reduced the watched resource Kind in cluster cache
- Update the controller to be 2 replicas for active/passive HA.

### Verification
Verified in a local cluster
```
make start 
```

To verify active/passive HA.
I checked the log for both pods, only one pod can initiate the synchronizer worker, the other one is block on getting the leader lock , e.g.`I0419 05:39:26.634182       1 leaderelection.go:248] attempting to acquire leader lease numaplane-system/numaplane-controller-lock...`

I also verified failover is successful, killed the leader pod and the other follower became leader and start to picking up sync tasks.


juliev0:
Is IKS requiring that we have High Availability/Leader Election? I don't believe we've really tested that out, have we? I don't want to update that part until we have.

I also don't truly understand why High Availability is always necessary (is it?). Kubernetes will just restart the Pod, and there will be a little downtime. For us, I'm not sure that that matters.

xdevxy:
> Is IKS requiring that we have High Availability/Leader Election? Has that been thoroughly tested? I don't want to update that part until we have.
> 
> I also don't truly understand why High Availability is always necessary (is it?). Kubernetes will just restart the Pod, and there will be a little downtime. For us, I'm not sure that that matters.
> 
> Update: Discussed offline here: https://intuit-teams.slack.com/archives/C05K82P3N9J/p1713373749594579 I suppose I'm okay with using it as long as we've thoroughly tested.

Hi @juliev0  what are the tests you are concerned? Thanks!

juliev0:
> > Is IKS requiring that we have High Availability/Leader Election? Has that been thoroughly tested? I don't want to update that part until we have.
> > I also don't truly understand why High Availability is always necessary (is it?). Kubernetes will just restart the Pod, and there will be a little downtime. For us, I'm not sure that that matters.
> > Update: Discussed offline here: https://intuit-teams.slack.com/archives/C05K82P3N9J/p1713373749594579 I suppose I'm okay with using it as long as we've thoroughly tested.
> 
> Hi @juliev0 what are the tests you are concerned? Can you elaborate? Thanks!

Thanks, Hao. I wasn't sure if 2 replicas have been tested to ensure that the follower replica is simply waiting and not doing anything. If I look at the code, I see just [Syncer.Start()](https://github.com/numaproj-labs/numaplane/blob/main/cmd/main.go#L115) being passed into the LeaderElection Runner. Is that right? Can the code below be executed by the follower?

xdevxy:
> > > Is IKS requiring that we have High Availability/Leader Election? Has that been thoroughly tested? I don't want to update that part until we have.
> > > I also don't truly understand why High Availability is always necessary (is it?). Kubernetes will just restart the Pod, and there will be a little downtime. For us, I'm not sure that that matters.
> > > Update: Discussed offline here: https://intuit-teams.slack.com/archives/C05K82P3N9J/p1713373749594579 I suppose I'm okay with using it as long as we've thoroughly tested.
> > 
> > 
> > Hi @juliev0 what are the tests you are concerned? Can you elaborate? Thanks!
> 
> Thanks, Hao. I wasn't sure if 2 replicas have been tested to ensure that the follower replica is simply waiting and not doing anything. If I look at the code, I see just [Syncer.Start()](https://github.com/numaproj-labs/numaplane/blob/main/cmd/main.go#L115) being passed into the LeaderElection Runner. Is that right? Can the code below be executed by the follower?

Yes, I verified when I pushed the change. I will update the PR description.

juliev0:
> Yes, I verified when I pushed the change. I updated the PR description, please see tests I did there. Thanks!

So, I guess my confusion is why only `syncer.Start()` needs to be in the `LeaderElectionRunner` and not the [code](https://github.com/numaproj-labs/numaplane/blob/main/cmd/main.go#L118-L137) which appears to start up the Reconciliation below, especially the call to `reconciler.SetupWithManager(mgr)`, which I thought would cause `Reconcile()` to be called. Can you help me understand the flow?

xdevxy:
> > Yes, I verified when I pushed the change. I updated the PR description, please see tests I did there. Thanks!
> 
> So, I guess my confusion is why only `syncer.Start()` needs to be in the `LeaderElectionRunner` and not the [code](https://github.com/numaproj-labs/numaplane/blob/main/cmd/main.go#L118-L137) which appears to start up the Reconciliation below, especially the call to `reconciler.SetupWithManager(mgr)`, which I thought would cause `Reconcile()` to be called. Can you help me understand the flow?

Yeah, sure. this is handled by [this](https://github.com/numaproj-labs/numaplane/blob/main/cmd/main.go#L58). basically when the leader lock is acquired by a replica, it will lock the k8s EndPoint resource of this replica set. The follower should no longer be able to reconcile.  On top of that, the syncer task should only be run in the leader.

juliev0:
> > > Yes, I verified when I pushed the change. I updated the PR description, please see tests I did there. Thanks!
> > 
> > 
> > So, I guess my confusion is why only `syncer.Start()` needs to be in the `LeaderElectionRunner` and not the [code](https://github.com/numaproj-labs/numaplane/blob/main/cmd/main.go#L118-L137) which appears to start up the Reconciliation below, especially the call to `reconciler.SetupWithManager(mgr)`, which I thought would cause `Reconcile()` to be called. Can you help me understand the flow?
> 
> Yeah, sure. this is handled by [this](https://github.com/numaproj-labs/numaplane/blob/main/cmd/main.go#L58). basically when the leader lock is acquired by a replica, it will lock the k8s EndPoint resource of this replica set. The follower should no longer be able to reconcile. On top of that, the syncer task should only be run in the leader.

Okay, thank you. Yes, this makes sense now.

juliev0 (approved):

PR #222, opened 2024-04-19
Moved external dependencies to the project

shubhamdixit863:
Yes

Fixes  https://github.com/numaproj-labs/numaplane/issues/101

### Modifications

Tests initially depended on external repositories. To resolve this, I have copied the necessary folders into our own repository where the tests reside and updated the repository paths in the tests to ensure they no longer depend on external sources.

### Verification
make test


internal/git/util_test.go
```diff
@@ -289,9 +289,9 @@ func Test_GetLatestManifests(t *testing.T) {
 				Spec: v1alpha1.GitSyncSpec{
 					GitSource: v1alpha1.GitSource{
 						GitLocation: v1alpha1.GitLocation{
-							RepoUrl:        "https://github.com/numaproj/numaflow.git",
-							TargetRevision: "main",
-							Path:           "config/namespace-install",
+							RepoUrl:        "https://github.com/numaproj-labs/numaplane.git",
+							TargetRevision: "controlledTestsPathEn",
```
juliev0, line 293: I guess you'll have to do a follow up commit unfortunately in which you need to change this to "main", huh? (sorry)

juliev0:
Thanks for doing this. I think there are still some tests which use `"https://github.com/numaproj-labs/numaplane-control-manifests.git"` - ideally, those should also be contained in this same repo I think as a best practice, but if this will take longer than half a day, I'm okay if you want to instead write up a second issue for that one, to be done in the future.

juliev0 (approved):

juliev0 (changes requested):
sorry, I approved initially, but then noticed that unit tests were failing

internal/git/util_test.go
shubhamdixit863: yes right once the code gets merged to main then i will make a follow up pr .Please dnt delete this branch post merge

shubhamdixit863:
> sorry, I approved initially, but then noticed that unit tests were failing

Now as this -issue -https://github.com/numaproj-labs/numaplane/issues/226 is resolved .We can merge this as well ,as test cases were failing because of the mentioned issue

juliev0 (approved):

juliev0:
Thanks @shubhamdixit863 ! I think we can consider the remaining tests related to "https://github.com/numaproj-labs/numaplane-control-manifests.git" to be pretty low priority for moving.

As a side note, for now there seem to be some unit tests that live in the [tests](https://github.com/numaproj-labs/numaplane/tree/main/tests) directory, and some that are in collocated `testdata` directories with the unit tests, like [this](https://github.com/numaproj-labs/numaplane/tree/main/internal/util/kubernetes/testdata). Perhaps this inconsistency is okay, and people can put them where they prefer? (My personal prefer is the collocated `testdata` directory since it's close by the test.) cc @xdevxy

PR #225, opened 2024-04-19
fix image repo; add initContainers to Deployment spec as empty array

juliev0:
### Modifications

Two modifications:
1. fix repo that we are pushing to in quay.io
2. Set `initContainers` to empty array in the Deployment spec, which makes it so that I can add an array-based patch to it in `numa-manifest-generator`. (I want to add an array-based patch, so that if we ever actually add an InitContainer in the open source side, we will simply append rather than overwriting.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ VERSION ?= latest
 BASE_VERSION := latest
 # Default cluster name where numaplane get deployed, update it as needed.
 CLUSTER_NAME ?= staging-usw2-k8s
-IMAGE_NAMESPACE ?= quay.io/numaproj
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -10,7 +10,7 @@ VERSION ?= latest
 BASE_VERSION := latest
 # Default cluster name where numaplane get deployed, update it as needed.
 CLUSTER_NAME ?= staging-usw2-k8s
-IMAGE_NAMESPACE ?= quay.io/numaproj
+IMAGE_NAMESPACE ?= quay.io/numaio
```
chandankumar4, line 13: Thanks for fixing this.

Issue #226, opened 2024-04-20
cloneRepo method in util.go gives Reference Not found error while cloning any branch other than main


shubhamdixit863:
**Describe the bug**  
When attempting to clone a repository branch using the `cloneRepo` method in `util.go`, the operation fails with a "Reference Not Found" error if the branch name is anything other than `main`. This issue seems related to how reference names are determined and used in the cloning process.

**To Reproduce**  
1. Call the `cloneRepo` method with a branch name other than `main`.
2. Observe that the method fails to clone the repository and throws a "Reference Not Found" error.
3. Note that no error occurs when the branch name is `main`.

**Expected behavior**  
The expected behavior is that the `cloneRepo` method should successfully clone any specified branch without errors. The method should dynamically determine the correct reference type and handle various branch names correctly.

**Additional context**  
The issue may be related to how the `GetReferenceName` and `GetRepoCloneOptions` functions handle the branch names and references, particularly for branches that do not follow typical naming conventions or when the branch name is not the default `main`. Further investigation into how these functions parse and use branch names could help identify the cause of the issue.

PR #227, opened 2024-04-22
Added reference name in clone options for git clone

shubhamdixit863:
Fixes  https://github.com/numaproj-labs/numaplane/issues/226

### Modifications

This PR modifies the `GetReferenceName` and `cloneRepo` functions to better handle branch names that are not the default `main`. Changes include:
- Updating the `GetReferenceName` function to more accurately determine and return appropriate `plumbing.ReferenceName` for any given branch or tag  name (Commit hash has been not included).
- Adjusting the `cloneRepo` method to use the updated reference name handling, ensuring that the cloning process supports all types of reference  names without throwing a "Reference Not Found" error.

### Verification

The changes were manually tested by  running  existing tests by `make test `.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ VERSION ?= latest
 BASE_VERSION := latest
 # Default cluster name where numaplane get deployed, update it as needed.
 CLUSTER_NAME ?= staging-usw2-k8s
-IMAGE_NAMESPACE ?= quay.io/numaproj
+IMAGE_NAMESPACE ?= quay.io/numaio
 IMAGE_FULL_PATH ?= $(IMAGE_NAMESPACE)/$(IMG):$(VERSION)
 
 ## Location to install dependencies to
--- a/Makefile
+++ b/Makefile
@@ -215,12 +215,12 @@ gitserver:
 # ref: https://stackoverflow.com/questions/33509194/command-to-delete-all-pods-in-all-kubernetes-namespaces
 .PHONY: e2e-test-clean
 e2e-test-clean:
-	$(KUBECTL) delete -n numaplane-e2e isbsvc --all || true
-	$(KUBECTL) delete -n numaplane-e2e pipelines --all || true
-	$(KUBECTL) delete -n numaplane-e2e cm --all || true
-	$(KUBECTL) delete -n numaplane-e2e secret --all || true
-	$(KUBECTL) delete -n numaplane-e2e all --all || true
-	$(KUBECTL) delete -n numaplane-system pod --all || true
+	$(KUBECTL) delete -n numaplane-e2e isbsvc --all
+	$(KUBECTL) delete -n numaplane-e2e pipelines --all
+	$(KUBECTL) delete -n numaplane-e2e cm --all
+	$(KUBECTL) delete -n numaplane-e2e secret --all
+	$(KUBECTL) delete -n numaplane-e2e all --all
+	$(KUBECTL) delete -n numaplane-system pod --all
 
 
 .PHONY: e2e-test-start
--- a/Makefile
+++ b/Makefile
@@ -215,12 +215,12 @@ gitserver:
 # ref: https://stackoverflow.com/questions/33509194/command-to-delete-all-pods-in-all-kubernetes-namespaces
 .PHONY: e2e-test-clean
 e2e-test-clean:
-	$(KUBECTL) delete -n numaplane-e2e isbsvc --all
-	$(KUBECTL) delete -n numaplane-e2e pipelines --all
-	$(KUBECTL) delete -n numaplane-e2e cm --all
-	$(KUBECTL) delete -n numaplane-e2e secret --all
-	$(KUBECTL) delete -n numaplane-e2e all --all
-	$(KUBECTL) delete -n numaplane-system pod --all
+	$(KUBECTL) delete -n numaplane-e2e isbsvc --all || true
+	$(KUBECTL) delete -n numaplane-e2e pipelines --all || true
+	$(KUBECTL) delete -n numaplane-e2e cm --all || true
+	$(KUBECTL) delete -n numaplane-e2e secret --all || true
+	$(KUBECTL) delete -n numaplane-e2e all --all || true
+	$(KUBECTL) delete -n numaplane-system pod --all || true
 
 
 .PHONY: e2e-test-start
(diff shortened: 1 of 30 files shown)


internal/util/git/git.go
```diff
@@ -81,8 +84,16 @@ func GetAuthMethod(ctx context.Context, repoCred *apiv1.RepoCredential, kubeClie
 	return auth, insecureSkipTLS, nil
 }
 
+// GetReferenceName dynamically determines the reference type and returns the appropriate plumbing.ReferenceName only for tags and branches
+func GetReferenceName(reference string) plumbing.ReferenceName {
+	if reference[0] == 'v' || reference[0] == 'V' {
+		return plumbing.NewTagReferenceName(reference)
```
juliev0, line 90: Do Tags always need to start with "v" or "V"? Should we maybe have the user specify whether it's a Tag or a Branch in the `GitSync` spec?

juliev0: If we decide that we should have this flexibility, I'm okay creating a follow up issue for that and just assuming for now.

juliev0: moreover, however, there can be branches that start with the letter "v"

shubhamdixit863: yeah i realised later this thing ,i am fixing this part .

juliev0: thanks

chandankumar4: Will it be better to have the targetRevision specified as: @juliev0 
- For tag - “tag/v1.0.0” 
- For branch - “branch_name”
?
Also do we need to support commit hash as well?

Makefile
```diff
@@ -215,12 +215,12 @@ gitserver:
 # ref: https://stackoverflow.com/questions/33509194/command-to-delete-all-pods-in-all-kubernetes-namespaces
 .PHONY: e2e-test-clean
 e2e-test-clean:
-	$(KUBECTL) delete -n numaplane-e2e isbsvc --all
-	$(KUBECTL) delete -n numaplane-e2e pipelines --all
-	$(KUBECTL) delete -n numaplane-e2e cm --all
-	$(KUBECTL) delete -n numaplane-e2e secret --all
-	$(KUBECTL) delete -n numaplane-e2e all --all
-	$(KUBECTL) delete -n numaplane-system pod --all
+	$(KUBECTL) delete -n numaplane-e2e isbsvc --all || true
```
xdevxy, line 218: Why do we need this?

internal/util/git/git.go
xdevxy: Is there any built in support in go git to be able to tell this? If not what @chandankumar4 is suggesting is better.

internal/git/util_test.go
```diff
@@ -428,8 +428,8 @@ AAAECl1AymWUHNdRiOu2r2dg97arF3S32bE5zcPTqynwyw50HAtto0bVGTAUATJhiDTjKa
 			Optional:        nil,
 		}},
 	}
-	repoUrL := "ssh://root@localhost:2222/var/www/git/test.git"
-	cloneOptions, err := gitshared.GetRepoCloneOptions(context.Background(), credential, client, repoUrL)
+	repoUrL := "ssh://root@localhost:2222/var/www/git/repo1.git"
+	cloneOptions, err := gitshared.GetRepoCloneOptions(context.Background(), credential, client, repoUrL, "master")
```
xdevxy, line 432: Can you add a test for git clone from tag as well?

internal/util/git/git.go
juliev0: So, @chandankumar4 are you suggesting that we add a field to the `GitSync` spec or no? If we do include commit hash, there may not be a good way to discern that from branch name just by name alone, right?

chandankumar4: I’m suggesting to use the existing field “spec.targetRevision” and for commit hash it can be like: “commit/<hash>” ?

chandankumar4 (commented):
I believe fetchUpdate() function should also need to be updated for referencing branch/tag/commit ?

internal/util/git/git.go
juliev0: Okay. I'll let @xdevxy weigh in on what she prefers.

xdevxy: @shubhamdixit863 can you follow the logic like https://github.com/src-d/go-git/blob/master/repository.go#L1362-L1387, so we don't make any assumption of how the revision should look like? Thanks!

shubhamdixit863: @xdevxy i went through the link ,but the problem is for it to work i need a cloned repository  reference.And for cloning i need to have the proper referenceName .What i can do is do a shallow clone in memory and check for the reference then clone in the temp directory

Makefile
shubhamdixit863: when e2e tests are run and there are no pods running for isbvc or secret is created it exits the test ,so to prevent exiting i have added this

juliev0: I'm not sure I understand. Is this when calling `make e2e-test-start`? (which calls this as a dependency) It seems that if you try to delete something that isn't there, then it would be okay and just keep going, no?

xdevxy (changes requested):
Please follow this suggestion https://github.com/numaproj-labs/numaplane/pull/227#discussion_r1575455110 if possible

shubhamdixit863:
Closing as raised another PR

internal/util/git/git.go
shubhamdixit863: > @shubhamdixit863 can you follow the logic like https://github.com/src-d/go-git/blob/master/repository.go#L1362-L1387, so we don't make any assumption of how the revision should look like? Thanks!

This is valid only when have repository reference (pointer) before even cloning we need to figure out the branch or tag

Makefile
shubhamdixit863: Removed this as discussed on slack chat

internal/git/util_test.go
shubhamdixit863: Yes added .

shubhamdixit863:
> I believe fetchUpdate() function should also need to be updated for referencing branch/tag/commit ?

i guess its not required if we are cloning the particular branch or tag the Head by default would point to the  that branch or tag and we can take pull or fetch without mentioning the branch or tagname in pull options

config/samples/numaplane.numaproj.io_v1alpha1_gitsync.yaml
```diff
@@ -4,9 +4,9 @@ metadata:
   name: gitsync-example
   namespace: numaplane-system
 spec:
-  path: "sample-pipeline"
```
xdevxy, line 7: Please keep the sample file as it is.

PR #236, opened 2024-04-24
chore: support for turning off auto heal

### LINKED ISSUE #235: Auto heal should be able to turn off
# Summary

What change needs making?

We should be able to turn off auto heal when the flag is off.

# Use Cases

When users want to intentionally overwrite the cluster state with auto heal off.

---

xdevxy:
Fixes #235 

### Modifications

This PR supports to turn off auto heal if it is configured so.

### Verification

Tested in a local cluster with `autoHealTimeIntervalMs` not set.  Auto heal is off so the syncing is skipped if no new changes.

### DIFF
--- a/internal/sync/syncer.go
+++ b/internal/sync/syncer.go
@@ -223,6 +223,13 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
 		numaLogger.Error(err, "error getting the global config")
 	}
 
+	autoHeal := true
+	interval := globalConfig.AutoHealTimeIntervalMs
+	// Check if auto heal is enabled
+	if interval == 0 {
+		autoHeal = false
+	}
+
 	numaLogger.SetLevel(globalConfig.LogLevel)
 
 	repo, err := git.CloneRepo(ctx, s.client, gitSync, globalConfig)
@@ -233,6 +240,22 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
 	if err != nil {
 		return fmt.Errorf("failed to get the manifest of key %q, %w", key, err)
 	}
+
+	namespacedName := types.NamespacedName{
+		Namespace: gitSync.Namespace,
+		Name:      gitSync.Name,
+	}
+	lastSyncedCommitHash, err := getCommitStatus(ctx, s.client, namespacedName, &numaLogger)
+	if err != nil {
+		return fmt.Errorf("failed to get the current commit status of key %q, %w", key, err)
+	}
+	// If auto heal is not enabled and the target commit hash is the same as last sync,
+	// skip the syncing.
+	if !autoHeal && lastSyncedCommitHash != nil && lastSyncedCommitHash.Hash == commitHash {
+		numaLogger.Info("Skip the syncing as there is no changes and auto heal turned off.")
+		return nil
+	}
+
 	uns, err := applyAnnotationAndNamespace(manifests, gitSyncName, gitSync.Spec.Destination.Namespace)
 	if err != nil {
 		return fmt.Errorf("failed to parse the manifest of key %q, %w", key, err)
@@ -243,11 +266,6 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
 	if syncState.Successful() {
 		synced = true
 		numaLogger.Info("GitSync object is successfully synced.")
-	}
-
-	namespacedName := types.NamespacedName{
-		Namespace: gitSync.Namespace,
-		Name:      gitSync.Name,
 	}
 
 	if !gitSync.GetDeletionTimestamp().IsZero() {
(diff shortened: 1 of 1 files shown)


internal/sync/syncer.go
```diff
@@ -233,6 +240,22 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
...
 		return fmt.Errorf("failed to get the manifest of key %q, %w", key, err)
 	}
+
+	namespacedName := types.NamespacedName{
+		Namespace: gitSync.Namespace,
+		Name:      gitSync.Name,
+	}
+	lastSyncedCommitHash, err := getCommitStatus(ctx, s.client, namespacedName, &numaLogger)
+	if err != nil {
+		return fmt.Errorf("failed to get the current commit status of key %q, %w", key, err)
```
juliev0, line 250: One thing to note is that we may generally only hit this error (if we do at all) because we've been rate limited by the K8S API Server. We could consider deciding that the only reason we were checking this to begin with was as an optimization, so we could theoretically continue. But I'm okay if maybe we want to just keep this as is.

internal/sync/syncer.go
```diff
@@ -233,6 +240,22 @@ func (s *Syncer) runOnce(ctx context.Context, key string, worker int) error {
...
+	}
+	// If auto heal is not enabled and the target commit hash is the same as last sync,
+	// skip the syncing.
+	if !autoHeal && lastSyncedCommitHash != nil && lastSyncedCommitHash.Hash == commitHash {
+		numaLogger.Info("Skip the syncing as there is no changes and auto heal turned off.")
```
juliev0, line 255:
```suggestion
		numaLogger.Info("Skip the syncing as there are no changes and auto heal is turned off.")
```

juliev0:
So, I'm good if it's a separate issue, but as we discussed offline, what do you think about our changing this:
```
	SyncTimeIntervalMs     int    `json:"syncTimeIntervalMs" mapstructure:"syncTimeIntervalMs"`
	AutoHealTimeIntervalMs int    `json:"autoHealTimeIntervalMs" mapstructure:"autoHealTimeIntervalMs"`
```

to:
```
	SyncTimeIntervalMs     int    `json:"syncTimeIntervalMs" mapstructure:"syncTimeIntervalMs"`
        AutoHealEnabled bool  `json:"autoHealEnabled" mapstructure:"autoHealEnabled"`
```

(I know originally, the bool was in there and I asked to remove it, but I think both of the other fields were as well, and it seems like we are really only using one time interval, right?)

internal/sync/syncer.go
xdevxy: Yeah, if we hit an error because of rate limiting, we will continue for the next round of syncing. So let's keep it simple and stick to the current logic.

xdevxy:
> So, I'm good if it's a separate issue, but as we discussed offline, what do you think about our changing this:
> 
> ```
> 	SyncTimeIntervalMs     int    `json:"syncTimeIntervalMs" mapstructure:"syncTimeIntervalMs"`
> 	AutoHealTimeIntervalMs int    `json:"autoHealTimeIntervalMs" mapstructure:"autoHealTimeIntervalMs"`
> ```
> 
> to:
> 
> ```
> 	SyncTimeIntervalMs     int    `json:"syncTimeIntervalMs" mapstructure:"syncTimeIntervalMs"`
>         AutoHealEnabled bool  `json:"autoHealEnabled" mapstructure:"autoHealEnabled"`
> ```
> 
> (I know originally, the bool was in there and I asked to remove it, but I think both of the other fields were as well, and it seems like we are really only using one time interval, right?)

Sounds good, let me do it in another PR. Thanks!

juliev0:
> > So, I'm good if it's a separate issue, but as we discussed offline, what do you think about our changing this:
> > ```
> > 	SyncTimeIntervalMs     int    `json:"syncTimeIntervalMs" mapstructure:"syncTimeIntervalMs"`
> > 	AutoHealTimeIntervalMs int    `json:"autoHealTimeIntervalMs" mapstructure:"autoHealTimeIntervalMs"`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to:
> > ```
> > 	SyncTimeIntervalMs     int    `json:"syncTimeIntervalMs" mapstructure:"syncTimeIntervalMs"`
> >         AutoHealEnabled bool  `json:"autoHealEnabled" mapstructure:"autoHealEnabled"`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (I know originally, the bool was in there and I asked to remove it, but I think both of the other fields were as well, and it seems like we are really only using one time interval, right?)
> 
> Sounds good, let me do it in another PR. Thanks!

thanks!

internal/sync/syncer.go
juliev0: sounds good

juliev0 (approved):

PR #238, opened 2024-04-24
test: e2e tests for Github CI (WIP)

dpadhiar:
Fixes #TODO

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,7 +224,7 @@ e2e-test-clean:
 
 
 .PHONY: e2e-test-start
-e2e-test-start: e2e-test-clean image
+e2e-test-start: e2e-test-clean image numaflow-crd
 	$(KUBECTL) apply -f tests/e2e/manifests/numaplane-ns.yaml
 	$(KUBECTL) apply -n numaplane-system -k ./tests/e2e-gitserver
 	$(KUBECTL) kustomize tests/e2e/manifests | sed 's/CLUSTER_NAME_VALUE/$(CLUSTER_NAME)/g' | sed 's@quay.io/numaproj/@$(IMAGE_NAMESPACE)/@' | sed 's/$(IMG):$(BASE_VERSION)/$(IMG):$(VERSION)/' | $(KUBECTL) apply -f -
@@ -236,3 +236,5 @@ test-%: e2e-test-start
 	go test -v -timeout 15m -count 1 --tags test -p 1 ./tests/$*
 	$(MAKE) e2e-test-clean
 
+numaflow-crd:
+	$(KUBECTL) apply -f numaflow_crd.yaml
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,6 +71,8 @@ jobs:
     name: E2E Tests
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    strategy:
+      fail-fast: false
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,6 +67,43 @@ jobs:
       - name: Test
         run: make test
 
+  e2e-tests:
+    name: E2E Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Restore go build cache
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/go-build
+          key: ${{ runner.os }}-go-build-v1-${{ hashFiles('**/go.mod') }}
+      - name: Setup Golang
+        uses: actions/setup-go@v4.0.0
+        with:
+          go-version: '1.21'
+      - name: Add bins to PATH
+        run: |
+          echo /home/runner/go/bin >> $GITHUB_PATH
+          echo /usr/local/bin >> $GITHUB_PATH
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -228,16 +229,14 @@ gitserver:
 # ref: https://stackoverflow.com/questions/33509194/command-to-delete-all-pods-in-all-kubernetes-namespaces
 .PHONY: e2e-test-clean
 e2e-test-clean:
-	$(KUBECTL) delete -n numaplane-e2e isbsvc --all
```
dpadhiar, line 231: removed as e2e test run on CI was hanging on deleting isbsvc / pipeline. unsure why as of now.

tests/fixtures/given.go
```diff
@@ -44,6 +46,11 @@ var (
 		Username: "root",
 		Password: "root",
 	}
+	author = &object.Signature{
```
dpadhiar, line 49: May not need author if we add a step to set git config in test.

Makefile
```diff
@@ -20,6 +20,7 @@ GIT_BRANCH=$(shell git rev-parse --symbolic-full-name --verify --quiet --abbrev-
 GIT_TAG=$(shell if [[ -z "`git status --porcelain`" ]]; then git describe --exact-match --tags HEAD 2>/dev/null; fi)
 GIT_TREE_STATE=$(shell if [[ -z "`git status --porcelain`" ]]; then echo "clean" ; else echo "dirty"; fi)
 
+NUMAFLOW_CRDS=$(shell kubectl get crd | grep -c 'numaflow')
```
juliev0, line 23:
```suggestion
NUMAFLOW_CRDS=$(shell kubectl get crd | grep -c 'numaflow.numaproj.io')
```

Makefile
juliev0: Are you deploying a Controller anywhere? If so, I know that would've been the original motivation for including these lines. 

If you do take this out, then the comment on lines 225-226 needs to be removed as well

dpadhiar: We are still deploying the controller in `TestNumaflowGitSync`. I was seeing an issue where the delete would hang on deleting the ISB or Pipeline which is odd cause locally those resources don't usually exist after all the tests have run. I will check tomorrow to see but otherwise I could remove the comment as well.

juliev0: If you're still deploying a Controller, then maybe you do need those lines? If so, if so is it possible to maybe add logging in the CI to figure out what's happening? The only thing I can think of is that the Controller somehow was previously in a state to add the Finalizer, but is no longer in a state to remove the Finalizer

dpadhiar: I can try something like that, from what I can see on the CI the controller is already long deleted but the pl/isbsvc still exist which is not the case when running locally. I could probably add a line to remove the finalizers if thats the case.

dpadhiar:
Converting to draft while testing different Makefile changes.

Makefile
juliev0: yeah, it seems okay if you want to remove the finalizers through patch or something

dpadhiar: That is the most likely what i'll do for now. Can get that added today.

xdevxy: is it possible for the test case to clean up the state with order to avoid the CI to do so. that may be easier to control?

xdevxy (commented):
LGTM.

Makefile
dpadhiar: Yes, in the test case there could a call to remove all files from the repo thus deleting the resources.

dpadhiar: Giving this approach a try now.

tests/e2e/functional_test.go
```diff
@@ -88,6 +88,11 @@ func (s *FunctionalSuite) TestNumaflowGitSync() {
 	w.Expect().ResourcesDontExist("numaflow.numaproj.io/v1alpha1", "pipelines", []string{"http-pipeline"})
 	w.Expect().CheckCommitStatus()
 
+	w.PushToGitRepo("numaflow/initial-commit", []string{"sample_pipeline.yaml", "isbsvc_jetstream.yaml"}, true).Wait(30 * time.Second)
+	w.Expect().ResourcesDontExist("numaflow.numaproj.io/v1alpha1", "pipelines", []string{"simple-pipeline"})
```
xdevxy, line 92: You may want to delete the pipeline first? but it may not matter.

xdevxy (approved):

xdevxy:
@juliev0 I'm good with the change, do you want to take another look?

tests/e2e/functional_test.go
juliev0: I think it's okay like this, as long as both are deleted before the controller

tests/e2e/functional_test.go
```diff
@@ -88,6 +88,11 @@ func (s *FunctionalSuite) TestNumaflowGitSync() {
 	w.Expect().ResourcesDontExist("numaflow.numaproj.io/v1alpha1", "pipelines", []string{"http-pipeline"})
 	w.Expect().CheckCommitStatus()
 
+	w.PushToGitRepo("numaflow/initial-commit", []string{"sample_pipeline.yaml", "isbsvc_jetstream.yaml"}, true).Wait(30 * time.Second)
```
juliev0, line 91:
```suggestion
	// Remove Pipeline and ISBService
	w.PushToGitRepo("numaflow/initial-commit", []string{"sample_pipeline.yaml", "isbsvc_jetstream.yaml"}, true).Wait(30 * time.Second)
```

PR #240, opened 2024-04-25
chore: refactor auto heal config

### LINKED ISSUE #239: Separate auto heal and syncing configs
# Summary

What change needs making?

As suggested [here](https://github.com/numaproj-labs/numaplane/pull/236#issuecomment-2074019674), we should have more accurate flag for auto healing and syncing intervals.

# Use Cases

When would you use this?
Auto heal should be configured as simple as enable or not, separating from syncing interval flag.

---

xdevxy:
Fixes #239 

### Modifications

Separate auto heal and syncing interval configs

### Verification

Verified in the local cluster.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -98,12 +98,7 @@ func main() {
 
 	numaLogger.SetLevel(config.LogLevel)
 
-	interval := config.AutoHealTimeIntervalMs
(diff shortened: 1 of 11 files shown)


juliev0 (approved):
thanks for doing that!

PR #241, opened 2024-04-25
Added reference name in clone options for git clone

shubhamdixit863:
Fixes https://github.com/numaproj-labs/numaplane/issues/226

Modifications
In this PR I have added a provision to checkout to the particular target revision as present in gitsync spec
1- If the target revision is branch then it directly checks out to the branch as mentioned [here](https://github.com/numaproj-labs/numaplane/blob/f051b9c9b47b0156203a19ff9e4b4c5b14bffc15/internal/git/util.go#L356) 

2- If target revision is tag then  it finds the first   branch which has that tag [here](https://github.com/numaproj-labs/numaplane/blob/f051b9c9b47b0156203a19ff9e4b4c5b14bffc15/internal/git/util.go#L299)  and then  checks out to that branch

3- If target revision is commit hash  then it finds the first branch which has that commit hash [here](https://github.com/numaproj-labs/numaplane/blob/f051b9c9b47b0156203a19ff9e4b4c5b14bffc15/internal/git/util.go#L400)  and then  checks out to that branch

4 - I have to comment out one test un resolvable target revision test  [here](https://github.com/numaproj-labs/numaplane/blob/f051b9c9b47b0156203a19ff9e4b4c5b14bffc15/internal/git/util_test.go#L265) ,which will be removed eventually as this test is no longer valid when we are checking out to the reference.

**Testing**

The changes were manually tested by running existing tests by make test .

### DIFF
--- a/internal/git/util.go
+++ b/internal/git/util.go
@@ -7,6 +7,9 @@ import (
 	"os"
 	"strconv"
 
+	"github.com/go-git/go-git/v5/plumbing/storer"
+	log "github.com/sirupsen/logrus"
+
 	"github.com/go-git/go-git/v5/config"
 
 	argoGit "github.com/argoproj/argo-cd/v2/util/git"
@@ -214,6 +217,22 @@ func getLocalRepoPath(gitSync *v1alpha1.GitSync) string {
 	}
 }
 
+func GetCurrentBranch(r *git.Repository) (string, error) {
+	// Get the HEAD reference to determine the current branch.
+	ref, err := r.Head()
+	if err != nil {
+		return "", fmt.Errorf("failed to get HEAD: %w", err)
+	}
+
+	// Check if the HEAD is a symbolic reference (which means it is pointing to a branch).
+	if ref.Name().IsBranch() {
+		return ref.Name().Short(), nil
+	}
+
+	// If HEAD is not a symbolic reference, it might be a detached HEAD.
+	return "", fmt.Errorf("HEAD is detached and not pointing to any branch")
+}
+
 // fetchUpdates fetches the remote branch and updates the local changes, returning nil if already up-to-date or an error otherwise.
 func fetchUpdates(ctx context.Context,
 	client k8sClient.Client,
@@ -226,7 +245,12 @@ func fetchUpdates(ctx context.Context,
 
 	credentials := gitShared.FindCredByUrl(gitSync.Spec.RepoUrl, globalConfig)
 
-	pullOptions, err := gitShared.GetRepoPullOptions(ctx, credentials, client, gitSync.Spec.RepoUrl)
+	branch, err := GetCurrentBranch(repo)
+	if err != nil {
+		log.Println("err", err)
+	}
+	log.Println("branch ---", branch)
+	pullOptions, err := gitShared.GetRepoPullOptions(ctx, credentials, client, gitSync.Spec.RepoUrl, branch)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


internal/git/util.go
```diff
@@ -245,14 +268,177 @@ func cloneRepo(ctx context.Context, gitSync *v1alpha1.GitSync, options *git.Clon
...
+			if openErr != nil {
+				return nil, fmt.Errorf("failed to open existing repo, err: %v", openErr)
+			}
+			return existingRepo, nil
+		}
+		return nil, err
+	}
+
+	// No need to fetch the references if  the target revision is already main or master (branches)
+	if gitSync.Spec.TargetRevision != "main" && gitSync.Spec.TargetRevision != "master" {
```
chandankumar4, line 284: Why we need to hardcode the "main" or "master" here, what I believe is git will clone the current active branch which can be anything, it doesn't tied with `main` or `master`

internal/git/util.go
```diff
@@ -245,14 +268,177 @@ func cloneRepo(ctx context.Context, gitSync *v1alpha1.GitSync, options *git.Clon
...
+		// fetch all references
+		err = fetchAll(r)
```
chandankumar4, line 286: why we need to fetch all brach of repo if we are only dealing with single branch/tag, as this will run in each sync which might cause performance issue if repo having more than 100 branches.

chandankumar4: I believe we can use the `targetReference` from gitsync spec to clone the particular branch/tag ? @xdevxy

internal/git/util.go
```diff
@@ -245,14 +268,177 @@ func cloneRepo(ctx context.Context, gitSync *v1alpha1.GitSync, options *git.Clon
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Perform checkout to the specified reference after a successful clone
+	if checkoutErr := checkoutRepo(r, gitSync.Spec.TargetRevision); checkoutErr != nil {
```
chandankumar4, line 293: Is it required to run in each reconsile? what if already checkedout?

shubhamdixit863: yeah that can be checked with current branch == targetRevsion.

internal/git/util.go
shubhamdixit863: as the default branch is either main or master ,and if we have given that in gitsync there is no point in doing the fetch .And why fetch is required because by default it doesnt gets the references so we need to do git fetch to get all the references and tags such that we can checkout on it

internal/git/util.go
shubhamdixit863: we have to fetch such that we can checkout to those branch or tag ,by default only the default reference is available and then if i do checkout it gives error detached head. Recursive process can be avoided by putting the  check if the repo is already checked to the  targetRevision branch or tag.

xdevxy (approved):

xdevxy (commented):
Let's address the comment in a follow up

internal/git/util.go
chandankumar4: okay

internal/git/util.go
chandankumar4: As per discussion, it's running for initial clone only, so it looks fine then.

xdevxy: I believe this code will be run each time when we reconcile. so we should avoid checking out if already does each time, right @shubhamdixit863 ?

shubhamdixit863: @xdevxy in this code above -```
r, err := git.PlainCloneContext(ctx, path, false, options)
	if err != nil {
		if errors.Is(err, git.ErrRepositoryAlreadyExists) {
			// Open the existing repo and return it.
			existingRepo, openErr := git.PlainOpen(path)
			if openErr != nil {
				return nil, fmt.Errorf("failed to open existing repo, err: %v", openErr)
			}
			return existingRepo, nil
		}
		return nil, err
	}
```
we are already checking if the repository is cloned or not  `if errors.Is(err, git.ErrRepositoryAlreadyExists) {` so if the repository is already cloned ,it means we have already checked it out ,so i guess it doesn't runs multiple times during reconcilation

PR #246, opened 2024-04-26
Enable credentials to be retrieved from files, not just from Kubernetes Secrets

### LINKED ISSUE #224: Github credentials from disk
# Summary

Currently, we have the ability to store our Github credentials in Kubernetes Secrets. We need another option to store them on local disk (generally it would be ephemeral storage), because this is considered more secure than in a K8S Secret.

(Plan would be that an InitContainer in the Controller would be able to fetch them and store them in a shared in-memory volume for Numaplane Controller to grab.)

Since a K8S Secret is essentially key/value pairs, can we similarly format this file as a json or yaml map of key/value pairs? How about we allow the user to do either? (Note I have just written the code [here](https://github.intuit.com/oss-analytics/addon-utilities/pull/7) to create a json file, which just contains a map in it).

Currently we have:
```
type SecretKeySelector struct {
	corev1.ObjectReference `json:",inline" mapstructure:",squash"` // for viper to correctly parse the config
	Key                    string                                  `json:"key" mapstructure:"key"`
	Optional               *bool                               `json:"optional,omitempty" mapstructure:"optional,omitempty"`
}
```

Can we update this struct (potentially break into multiple or keep as one) so that it can alternatively use a file path as opposed to a K8S Secret? (so basically the `corev1.ObjectReference` would be optional and could alternatively be a file path which could be json or yaml)

---

juliev0:
Fixes #224 

### Modifications

The RepoCredentials (both HTTP and SSH) can now either come from a K8S Secret or from a file, as configured in the Configmap.

### Verification

This has been tested both locally as well as on an actual Kubernetes try cluster with an InitContainer writing an HTTP credential to a json file and the Numaplane Controller reading it in from shared volume.

Locally, I have written unit tests to verify getting credentials from files.


pkg/apis/numaplane/v1alpha1/shared_types_test.go
```diff
@@ -0,0 +1,88 @@
...
+limitations under the License.
+*/
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_GetSecretValue(t *testing.T) {
```
xdevxy, line 25: nit: `Test_GetSecretValueFromFile`?

internal/controller/config/config_test.go
```diff
@@ -44,21 +44,21 @@ func TestLoadConfigMatchValues(t *testing.T) {
 
 	assert.NotNil(t, config.RepoCredentials[0].HTTPCredential, "HTTPCredential for numaproj-labs is missing")
 	assert.Equal(t, "exampleUser", config.RepoCredentials[0].HTTPCredential.Username, "Username for HTTPCredential of numaproj-labs does not match")
-	assert.Equal(t, "http-creds", config.RepoCredentials[0].HTTPCredential.Password.Name, "Password Name for HTTPCredential of numaproj-labs does not match")
-	assert.Equal(t, "password", config.RepoCredentials[0].HTTPCredential.Password.Key, "Password Key for HTTPCredential of numaproj-labs does not match")
-	assert.Equal(t, "numaplane-controller", config.RepoCredentials[0].HTTPCredential.Password.Namespace, "Kubernetes namespace for password doesn't match")
+	assert.Equal(t, "http-creds", config.RepoCredentials[0].HTTPCredential.Password.FromKubernetesSecret.Name, "Password Name for HTTPCredential of numaproj-labs does not match")
```
xdevxy, line 47: Should we add a test to read from file as well?

internal/git/util_test.go
```diff
@@ -437,11 +437,10 @@ AAAECl1AymWUHNdRiOu2r2dg97arF3S32bE5zcPTqynwyw50HAtto0bVGTAUATJhiDTjKa
 	assert.Nil(t, err)
 
 	credential := &v1alpha1.RepoCredential{
-		SSHCredential: &v1alpha1.SSHCredential{SSHKey: v1alpha1.SecretKeySelector{
+		SSHCredential: &v1alpha1.SSHCredential{SSHKey: v1alpha1.SecretSource{FromKubernetesSecret: &v1alpha1.SecretKeySelector{
```
xdevxy, line 440: Do you think such a test to read the secret from file is needed here?

xdevxy (commented):
LGTM, mostly is about how can we test this?

internal/git/util_test.go
juliev0: yeah, I could add that

internal/controller/config/config_test.go
juliev0: sure

juliev0: incorporated into this test

internal/git/util_test.go
juliev0: just added below

juliev0:
addressed all comments

xdevxy (approved):

PR #250, opened 2024-04-26
Update ClusterRole with minimal permissions

### LINKED ISSUE #243: Need to reevaluate ClusterRole access
**Describe the bug**

We may need to be able to do:
```
- apiGroups:
  - ""
  resources:
  - secrets
  - namespaces
  verbs:
  - 'get'
  - 'list'
  - 'watch'
```

I'm not sure offhand about what we need "namespaces" for, but the code was failing without it (see [this](https://github.com/numaproj-labs/numaplane/issues/242) issue).

Need to also see if anything else is missing.

---

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

xdevxy:
Fixes #243 

### Modifications
This PR updates ClusterRole with minimal permissions if without namespace rolebing.

### Verification
Tested in local cluster

### DIFF
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -372,6 +372,15 @@ rules:
   verbs:
   - '*'
 - apiGroups:
+  - ""
+  resources:
+  - secrets
+  - namespaces
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - apps
   resources:
   - deployments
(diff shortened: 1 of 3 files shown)


config/install.yaml
```diff
@@ -419,7 +428,7 @@ data:
     syncTimeIntervalMs: 30000
     autoHealEnabled: true
     includedResources: "group=apps,kind=Deployment;\
-    group=,kind=ConfigMap;group=,kind=Secret;group=,kind=ServiceAccount;group=,kind=Namespace;\
```
juliev0, line 422: so, these are the Resources that gitops-engine keeps track of that we're syncing, right? I think we only need `Secret` for ourselves to be able to read in Git credentials, so it doesn't need to be here, does it?

xdevxy: I guess we can keep it for open source version in case pipeline needs secret as well

juliev0: If we include `Secret`, then maybe we should also include `Role` and `RoleBinding` in case their Pipeline Pods' Service Accounts need to be bound to Roles.

juliev0: Also, the verbs for the role need to be updated to “*” if we include Secrets and Namespaces here, right?

also, what about Service kind?

xdevxy: yeah, in that case, we should allow write permission for `Secret`. so let me just take out `Secret` as it really doesn't matter, the user can patch on it.

juliev0 (approved):

PR #253, opened 2024-04-29
test: improving autohealing e2e test

### LINKED ISSUE #247: E2E test for auto heal is off
E2E test when auto heal is disabled, the manifest is not auto healed to the git target again.

dpadhiar:
Fixes #247 

### Modifications

Improves autohealing test by adding the option to enable or disable autohealing through the Numaplane controller configmap. This allows us to test that resources are not healed to the Git repo state.

### Verification

Ran `make test-e2e` locally.

### DIFF
--- a/tests/e2e/functional_test.go
+++ b/tests/e2e/functional_test.go
@@ -178,7 +178,7 @@ func (s *FunctionalSuite) TestAutoHealing() {
 	w.Expect().VerifyResourceState("apps/v1", "deployments", "test-deploy", "spec", "replicas", 4)
 
 	// reenable autohealing for test cases left to run
-	w.UpdateConfig(true)
+	w.UpdateAutoHealConfig(true)
 }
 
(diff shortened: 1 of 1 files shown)


tests/fixtures/when.go
```diff
@@ -226,6 +227,42 @@ func (w *When) ModifyResource(apiVersion, resourceType, resource, patch string)
 	return w
 }
 
+// update Numaplane controller configmap
+// configurable currently: autohealing
+func (w *When) UpdateConfig(autoHealEnabled bool) *When {
```
xdevxy, line 232: nit: name it to `UpdateAutoHealConfig`

xdevxy (approved):
Just one nit

tests/fixtures/when.go
dpadhiar: Can do.

Issue #88, opened 2024-02-13
Use of persistence storage for cloning repository


chandankumar4:
**What happened**:
- For cloning the repo we are using some path inside kubernetes pod for cloning the repo, the problem with that is whenever pod restart due to some reason then it has to clone all repos again.

**What you expected to happen**:
- Using persistence repo we don't need to clone repo again and again in case of restart, we can just fetch the new changes.

**Anything else we need to know?**:

PR #257, opened 2024-04-30
Persistent storage

shubhamdixit863:
Yes

Fixes https://github.com/numaproj-labs/numaplane/issues/88

### Modifications

Added a persistent volume and persistent volume claim to the GitSync controller to map the path with the cluster node path or host path. Additionally, modified the git clone process to utilize this persistent volume claim path for cloning repositories.

### Verification

The code has been tested in a local Kind cluster, where the cloned repository is visible in the cluster node's host path. Since the Kind cluster runs inside a Docker container, executing the `docker exec -it` command with shell access allows entry into the host path specified in the PV YAML file.

### DIFF
--- a/config/manager/persistent-volume-claim.yaml
+++ b/config/manager/persistent-volume-claim.yaml
@@ -1,9 +1,9 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: repo-pvc
+  name: numaplane-repo-pvc
 spec:
-  storageClassName: manual
+  storageClassName: standard
   accessModes:
     - ReadWriteOnce
   resources:
--- a/config/manager/persistent-volume-claim.yaml
+++ b/config/manager/persistent-volume-claim.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: repo-pvc
+spec:
+  storageClassName: manual
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 5Gi
--- a/internal/git/util.go
+++ b/internal/git/util.go
@@ -214,13 +214,11 @@ func isRootDir(path string) bool {
 // by default it will use /tmp as base directory
 // unless LOCAL_REPO_PATH env is set.
 func getLocalRepoPath(gitSync *v1alpha1.GitSync) string {
-	baseDir := os.Getenv("LOCAL_REPO_PATH")
+	// baseDir is persistent volume path on the cluster node
+	// TODO // should we get it from config variable ?
+	baseDir := "/etc/gitrepo"
 	repoUrl := strconv.FormatUint(xxhash.Sum64([]byte(gitSync.Spec.RepoUrl)), 16)
-	if baseDir != "" {
-		return fmt.Sprintf("%s/%s/%s", baseDir, gitSync.Name, repoUrl)
-	} else {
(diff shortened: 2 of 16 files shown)


config/manager/persistent-volume-claim.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: repo-pvc
+spec:
+  storageClassName: manual
+  accessModes:
+    - ReadWriteOnce
```
juliev0, line 8: I wonder if we need `ReadWriteMany`? I see "ReadWriteOnce" means "the volume can be mounted as read-write by a single node. ReadWriteOnce access mode still can allow multiple pods to access the volume when the pods are running on the same node." We will have one Leader Controller Pod, and one Follower Controller Pod, and they can be on different nodes.

internal/git/util.go
```diff
@@ -214,13 +214,11 @@ func isRootDir(path string) bool {
 // by default it will use /tmp as base directory
 // unless LOCAL_REPO_PATH env is set.
 func getLocalRepoPath(gitSync *v1alpha1.GitSync) string {
-	baseDir := os.Getenv("LOCAL_REPO_PATH")
+	// baseDir is persistent volume path on the cluster node
+	// TODO // should we get it from config variable ?
```
xdevxy, line 218: yes, please add a config. Thanks!

config/manager/persistent-volume-claim.yaml
xdevxy: It seems fine as we should only read/write from the leader at a time?

xdevxy (commented):
Have you verified in the local cluster the scenario that delete the leader pod and see if the new leader can continue the syncing without cloning again? Thanks!

config/manager/persistent-volume-claim.yaml
juliev0: I'm not sure? Is it?

juliev0:
> Have you verified in the local cluster the scenario that delete the leader pod and see if the new leader can continue the syncing without cloning again? Thanks!

Do you mean regarding the `ReadWriteOnce` setting? But in order to truly test that, the two Pods would have to be on different Nodes, right?

config/manager/persistent-volume-claim.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: repo-pvc
+spec:
+  storageClassName: manual
```
chandankumar4, line 6: will there be already a storage class with name as `manual` ? @juliev0

config/manager/persistent-volume-claim.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: repo-pvc
```
chandankumar4, line 4: probably need to rename it something like: `numaplane-repo-pvc` or `numaplane-pvc`

config/manager/persistent-volume-claim.yaml
chandankumar4: Oh, I found an open issue related to it [here](https://github.com/kubernetes/website/issues/30912)

xdevxy: https://github.com/kubernetes/website/issues/30912#issuecomment-1828679721

PR #259, opened 2024-04-30
Sync with git config file

afugazzotto:
Fixes https://github.com/numaproj-labs/numaplane/issues/262

### Modifications

Added a function to load the gitconfig file and apply URL mapping and Authorization config options if available for cloning and fetching operations.

### Verification

Tested manually by running the controller on K8s cluster with and without the URL mappings and Authorization extraheader in gitconfig file.

### DIFF
--- a/internal/util/git/git.go
+++ b/internal/util/git/git.go
@@ -132,7 +132,7 @@ func GetRepoCloneOptions(ctx context.Context, repoCred *apiv1.RepoCredential, ku
 // GetRepoPullOptions creates git.PullOptions for pull updates from a repo with HTTP, SSH, or TLS credentials from Kubernetes secrets.
 func GetRepoPullOptions(ctx context.Context, repoCred *apiv1.RepoCredential, kubeClient k8sClient.Client, repoUrl string, refName string) (*git.PullOptions, error) {
 	// check to ensure proper repository url is passed
-	_, err := transport.NewEndpoint(repoUrl)
+	remoteURL, err := transport.NewEndpoint(repoUrl)
 	if err != nil {
 		return nil, fmt.Errorf("invalid repository URL: %w", err)
 	}
@@ -146,6 +146,7 @@ func GetRepoPullOptions(ctx context.Context, repoCred *apiv1.RepoCredential, kub
 		InsecureSkipTLS: skipTls,
 		RemoteName:      "origin",
 		ReferenceName:   plumbing.NewBranchReferenceName(refName),
+		RemoteURL:       remoteURL.String(),
 	}, nil
 }
 
@@ -157,6 +158,10 @@ func GetRepoFetchOptions(
 	kubeClient k8sClient.Client,
 	repoUrl string,
 ) (*git.FetchOptions, error) {
+	remoteURL, err := transport.NewEndpoint(repoUrl)
+	if err != nil {
+		return nil, fmt.Errorf("invalid repository URL: %w", err)
+	}
 
 	method, skipTls, err := GetAuthMethod(ctx, repoCred, kubeClient, repoUrl)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


xdevxy:
> Fixes
> 
> Fixed errors logging caller. Depth needed to be adjusted because showing the logger as caller.
> 
> ### Modifications
> Added a function to load the gitconfig file and apply URL mapping and Authorization config options if available for cloning and fetching operations.
> 
> ### Verification
> Tested manually by running the controller on K8s cluster with and without the URL mappings and Authorization extraheader in gitconfig file.

Do you mind creating a feature tracking issue for it? Thanks!

internal/util/git/git.go
```diff
@@ -174,3 +193,70 @@ func NormalizeGitUrl(gitUrl string) string {
...
+			}
+		}
+	case git.FetchOptions, git.PullOptions:
+		httpSubSecKey = fmt.Sprintf("%s://%s", rURL.Scheme, rURL.Host)
+	}
+
+	var authzHeader *AuthorizationHeader
+	if httpSection := gitConfig.Raw.Section("http"); httpSection != nil {
+		if subSection := httpSection.Subsection(httpSubSecKey); subSection != nil {
+			if option := subSection.Option("extraheader"); strings.HasPrefix(option, "Authorization:") {
```
juliev0, line 242: just out of curiosity, do you think in the future there will be other types of headers we need to handle?

juliev0: (not at Intuit per se, but in general)

juliev0: and maybe it's really a general question about "things that could go in a .gitconfig file"

internal/util/git/git.go
```diff
@@ -174,3 +193,70 @@ func NormalizeGitUrl(gitUrl string) string {
...
+					return fmt.Errorf("invalid URL '%s' in git config: %v", k, err)
+				}
+
+				any(options).(*git.CloneOptions).URL = fmt.Sprintf("%s%s", k, rURL.Path)
+				httpSubSecKey = fmt.Sprintf("%s://%s", keyURL.Scheme, keyURL.Host)
+				break
+			}
+		}
+	case git.FetchOptions, git.PullOptions:
+		httpSubSecKey = fmt.Sprintf("%s://%s", rURL.Scheme, rURL.Host)
```
juliev0, line 236: I see `git.FetchOptions` and `git.PullOptions` both have a `remoteURL` field in them - want to verify we don't need to set that, do we?

internal/util/git/git.go
juliev0: sounds good, thanks

internal/util/git/git.go
afugazzotto: With my initial tests, I was passing the remote URL for the fetch options update. After making adjustments due to merge conflicts, I think it would be beneficial to set the RemoteURL field. I'm working on testing that now.

juliev0 (approved):

internal/util/git/git.go
```diff
@@ -138,6 +158,10 @@ func GetRepoFetchOptions(
 	kubeClient k8sClient.Client,
 	repoUrl string,
 ) (*git.FetchOptions, error) {
+	remoteURL, err := transport.NewEndpoint(repoUrl)
```
xdevxy, line 161: Good catch! Thanks!

internal/util/git/git_test.go
```diff
@@ -370,3 +374,119 @@ func TestHTTPAuthMethodFile(t *testing.T) {
 	assert.Equal(t, "someuser", basicAuth.Username)
 	assert.Equal(t, "my-password", basicAuth.Password)
 }
+
+// NOTE: FetchOptions and PullOptions would be tested similarly except that the URL field would be RemoteURL instead.
```
xdevxy, line 378: I guess we can run the same test cases with different options. but it is not necessary I assume.

xdevxy (approved):

Issue #189, opened 2024-04-02
Configmap variable removal not resetting value immediately


afugazzotto:
**Describe the bug**
When removing the LogLevel variable from configmap, it takes 1 or more iterations for the variable to be reset to zero. This could impact also other variables (not only the LogLevel).

**To Reproduce**

1. add logLevel with a valid non-zero value (ex: 4 = debug) to the configmap using `kubectl edit`;
2. the logger should start showing appropriate logs based on level (ex: debug messages);
3. while tailing the logs, use `kubectl edit` to remove the logLevel variable from the configmap;
4. the variable value will not change to zero for 1 or more cycles, then it will change to zero.

**Expected behavior**
When removing the variable from configmap, the value should be applied immediately.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

juliev0:
As far as this bug, I'm kind of wondering if it's reasonable for there to be a delay when re-reading the ConfigMap? And that if someone wants immediacy then they should restart the Pod? 

But it's true that we should revisit our general strategy when it comes to re-ingesting the ConfigMap and see if there are improvements that can be made.

I think by "iteration"/"cycle", you mean where the re-ingestion is happening [here](https://github.com/numaproj-labs/numaplane/blob/main/internal/sync/syncer.go#L221-L226) each time `runOnce()` is called? The only way to have immediacy would be for the Log functions - `Info()`, `Debug()` -  to reference directly the global config each time when logging. 

Background: The way we wrote the ConfigMap logic, we wanted to have thread safety because there's a separate [thread](https://github.com/numaproj-labs/numaplane/blob/main/internal/controller/config/config.go#L74) which does the write. 
There are 2 potential issues: 1. let's say a reader is only interested in one field - well, at least that one field that they read should be safe to read. 2. let's say a reader is interested in more than one field - that combination of fields should also be atomically correct. 

I guess the current limitation is that we need to make sure that every long-running thread needs to proactively call `controllerConfig.GetConfigManagerInstance().GetConfig()` each time it wants the latest and greatest info. 

[...]

PR #265, opened 2024-05-01
update context when loglevel is updated

juliev0:
### Motivation

We are using the [context to get the Logger](https://github.com/numaproj-labs/numaplane/blob/v0.1.2/internal/sync/syncer.go#L117), but we don't ever update the context to reflect that the Logger changed.

### Modifications

I added a Base Logger for the application, which is essentially the original Logger but with the Log level periodically updated based on the Config. 

Long living goroutines periodically refresh the Base Logger with the current log level and get a copy of it, add any Values they want logged, and then update the context to include that Logger, so downstream users will log those same values automatically. 

### Verification

Updated logLevel from `info` to `debug` level back and forth, and witnessed the log lines change. Verified for both calls in `syncer.runOnce()`, calls from functions called downstream of `syncer.runOnce()` for which `context` got passed as argument, and from `Reconcile()`.

### Remaining work

1. We're still not updating the LogrLogger log level passed to the underlying cache in gitops-engine. That is constructed [here](https://github.com/numaproj-labs/numaplane/blob/main/internal/sync/cache.go#L184) and never gets updated. This is the one place where the logging is still not changing if you look at the log lines.
2. [This](https://github.com/numaproj-labs/numaplane/issues/189) issue still exists. It takes a couple of cycles for the log level to change where it does work.

### DIFF
--- a/internal/util/logger/logger.go
+++ b/internal/util/logger/logger.go
@@ -58,7 +58,9 @@ var logrVerbosityToZerologLevelMap = map[int]zerolog.Level{
 	VerboseLevel: -2,
 }
 
-type loggerKey struct{}
+var (
+	loggerKey = "logger"
+)
 
 // NumaLogger is the struct containing a pointer to a logr.Logger instance.
 type NumaLogger struct {
@@ -74,7 +76,7 @@ type LogSink struct {
 
 // New returns a new NumaLogger with a logr.Logger instance with a
 // default setup for zerolog, os.Stdout writer, and info level.
-func New() NumaLogger {
+func New() *NumaLogger {
 	w := io.Writer(os.Stdout)
 
 	lvlStr := os.Getenv(common.EnvLogLevel)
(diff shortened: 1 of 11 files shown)


internal/util/logger/logger.go
```diff
@@ -74,22 +75,24 @@ type LogSink struct {
 
 // New returns a new NumaLogger with a logr.Logger instance with a
 // default setup for zerolog, os.Stdout writer, and info level.
-func New() NumaLogger {
+func New() *NumaLogger {
 	w := io.Writer(os.Stdout)
 
-	lvlStr := os.Getenv(common.EnvLogLevel)
```
juliev0, line 80: Is there any reason why we need this environment variable? Seems better to just have one place (the ConfigMap) for storing it

internal/util/logger/logger.go
```diff
@@ -120,116 +123,126 @@ func newNumaLogger(writer *io.Writer, level *int) NumaLogger {
...
-	return NumaLogger{&ll}
+	return &NumaLogger{&ll, lvl}
 }
 
 // WithLogger returns a copy of parent context in which the
 // value associated with logger key is the supplied logger.
-func WithLogger(ctx context.Context, logger NumaLogger) context.Context {
-	return context.WithValue(ctx, loggerKey{}, logger)
+func WithLogger(ctx context.Context, logger *NumaLogger) context.Context {
+	return context.WithValue(ctx, loggerKey{}, logger.DeepCopy())
```
juliev0, line 132: It seemed to not successfully save the logger into the context when it was copying the value for some reason so I modified it to a pointer

afugazzotto (approved):

Issue #270, opened 2024-05-02
Git server: unable to push to repos that don't require authentication


dpadhiar:
**Describe the bug**

While writing an e2e test case for public repos that do not require authentication, I found that I was unable to push to the repo after cloning and adding files to commit. This was occurring using the e2e test suite as well as manually cloning and attempting to push to the repo by port-forwarding the Git server pod.

**To Reproduce**

1. Port-forward localgitserver pod
2. Clone a `gitopen` repo from the pod to your machine
3. Create any file and add it to a commit to push
4. Push commit and observe 403 Forbidden error

**Expected behavior**
Pushing to this repo should work just like any other public repo.

**Screenshots**

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.18.6]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

PR #271, opened 2024-05-02
fixed public git repo serve for apache server

shubhamdixit863:
Yes

Fixes  https://github.com/numaproj-labs/numaplane/issues/270

### Modifications
I had to modify the Apache server configuration for LocalGitServer because there was an issue resolving public Git URLs. The location matching for `/gitopen` was getting redirected to '/git', which requires authentication, so I changed `/gitopen` to `/public-git`. 

### Verification

Tests running on local cluster and unit tests

### DIFF
--- a/tests/gitserver/Dockerfile
+++ b/tests/gitserver/Dockerfile
@@ -46,7 +46,7 @@ RUN for i in $(seq 1 $REPO_COUNT); do \
     chmod +x "/var/www/git/repo$i.git/hooks/post-receive"; \
     done
 
-#creating repositories which donot require authentication and can be accessed from http://localhost:8080/public-git
+#creating repositories which don't require authentication and can be accessed from http://localhost:8080/public-git
 RUN for i in $(seq 1 $REPO_COUNT); do \
     mkdir -p "/var/www/public-git/repo$i.git" && \
     git init --bare "/var/www/public-git/repo$i.git" && \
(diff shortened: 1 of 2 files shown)


tests/gitserver/Dockerfile
```diff
@@ -46,14 +46,14 @@ RUN for i in $(seq 1 $REPO_COUNT); do \
     chmod +x "/var/www/git/repo$i.git/hooks/post-receive"; \
     done
 
-#creating repositories which donot require authentication and can be accessed from http://localhost:8080/gitopen
+#creating repositories which donot require authentication and can be accessed from http://localhost:8080/public-git
```
dpadhiar, line 49:
```suggestion
#creating repositories which don't require authentication and can be accessed from http://localhost:8080/public-git
```

tests/gitserver/Dockerfile
```diff
@@ -125,17 +125,17 @@ RUN for i in $(seq 1 $REPO_COUNT); do \
 # Creating a non bare git repo  which donot require  authentication with working directory as per the RepoCount variable and pushing the changes
```
dpadhiar, line 125:
```suggestion
# Creating a non bare git repo which don't require authentication with working directory as per the RepoCount variable and pushing the changes
```

dpadhiar:
Were you able to port-forward the Git server pod and clone it / push to it as well?

PR #287, opened 2024-05-09
Pvc claim with efs storage

shubhamdixit863:
Yes

Fixes (https://github.intuit.com/oss-analytics/numaproj/issues/50)

### Modifications

Added storage class with efs  and created new pvc yaml using the storage class created for volume claim

### Verification

Yet to be done

### DIFF
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -533,4 +533,4 @@ spec:
         name: config-volume
       - name: repo-volume
         persistentVolumeClaim:
-          claimName: numaplane-repo-pvc
+          claimName: numaplane-efs-claim-data-1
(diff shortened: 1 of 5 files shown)


juliev0:
Sorry to chime in late on all of this, but I'm thinking that anything which will only work in some types of users' environments and not others is something we shouldn't put here but rather just in the numa-manifest-generator. Is that the case for this? Is that the case for the previous PR that went in?

xdevxy:
> Sorry to chime in late on all of this, but I'm thinking that anything which will only work in some types of users' environments and not others is something we shouldn't put here but rather just in the numa-manifest-generator. Is that the case for this? Is that the case for the previous PR that went in?

Right, this change should be in `numa-manifest-generator`.

PR #298, opened 2024-05-14
test: improving turnaround time on e2e tests

### LINKED ISSUE #292: Improve turnaround time on e2e tests
# Summary

Probably want to create multiple Test Suites that can run in parallel in a matrix.

Potentially reduce automatic syncing interval. 

---

dpadhiar:
Fixes #292 

### Modifications

Changes the `syncIntervalTime` in the e2e test configs to 10 seconds. This allows us to decrease the waits that occur after Git or config changes as the controller reconciles the changes that much faster.

### Verification

Running `make-x-e2e` for all suites locally and on CI.


dpadhiar:
`AutoHealing` test is showing that it can be flakey. A re-run usually shows that it will pass but unsure what root cause is for now.

juliev0 (approved):

PR #302, opened 2024-05-20
Added NumaflowControllerRollout and ISBServiceRollout CRDs and Controllers

afugazzotto:


juliev0 (approved):

xdevxy (approved):

config/rbac/isbservicerollout_editor_role.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: numaplane
+    app.kubernetes.io/managed-by: kustomize
+  name: isbservicerollout-editor-role
+rules:
+- apiGroups:
+  - numaplane.numaproj.io.github.com.numaproj-labs
```
whynowy, line 11: Are we still using this group?

xdevxy: Good catch! No, we should be using `numaplane.numaproj.io`. This could be an issue with kubebuilder.

PR #306, opened 2024-05-20
Update spec

juliev0:
getting shared code in so Tony can use it


afugazzotto (approved):
Based on our conversation, we will change the spec to not include k8s api details and metadata and use spec as RawExtension.

PR #307, opened 2024-05-21
Add configmap to store numaflow controller spec

xdevxy:
### Modifications

This PR adds controller definitions config to store the specs for numaflow controller in different versions. This config is immutable as the definition should not be altered accidentally. When new definitions come in, we should recreate this config.

### Verification

Unit test and `make start`


config/manager/controller-manager.yaml
```diff
@@ -83,9 +83,14 @@ spec:
...
+            mountPath: /etc/numarollout
       serviceAccountName: numaplane-sa
       terminationGracePeriodSeconds: 10
       volumes:
         - name: config-volume
           configMap:
             name: numaplane-controller-config
+        - name: numarollout-volume
+          configMap:
+            name: numarollout-config
```
juliev0, line 96: yeah, if we volume mount the file, you're right that it probably makes the most sense to have just a single file with one key for each version.

(An alternative we could potentially weigh in the future, if for some reason it's better to have multiple files, is to make a call to K8S API Server to Get all of the ConfigMaps with a particular label or something. I'm just thinking "what's the easiest way from the `numa-manifest-generator` side to create everything?" But maybe one file still works there as well...hmm)

juliev0 (approved):

PR #313, opened 2024-05-22
Add generic Status to all controllers

afugazzotto:
Example: `kubectl describe pipelinerollout my-pipeline`
image


.gitignore
```diff
@@ -30,4 +30,6 @@ Dockerfile.cross
 # tmp directories built for e2e tests
 tests/e2e/local
 tests/kustomize-e2e/local
-tests/helm-e2e/local
\ No newline at end of file
+tests/helm-e2e/local
+
+config/crd/bases/_.yaml
```
juliev0, line 35: good solution for now at least :)

config/crd/bases/numaplane.numaproj.io_gitsyncs.yaml
```diff
@@ -203,11 +203,10 @@ spec:
                   type: object
                 type: array
               message:
-                description: Message is added if there's a failure
+                description: Message is added if Phase is PhaseFailed.
```
juliev0, line 206: typo?

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -0,0 +1,176 @@
...
+
+// Status is a common structure which can be used for Status field.
+type Status struct {
+	// Conditions are the latest available observations of a resource's current state.
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+
+	// Phase indicates the status of the underlying CR.
```
juliev0, line 54: since this also applies to `GitSync` maybe we shouldn't say anything about underlying CR here? maybe just generically it really represents the health of this, whatever it means

internal/controller/pipelinerollout_controller.go
```diff
@@ -158,15 +156,16 @@ func (r *PipelineRolloutReconciler) reconcile(ctx context.Context, pipelineRollo
 		return err
```
juliev0, line 156: how about the "todo" I have above?

pkg/apis/numaplane/v1alpha1/status_types.go
```diff
@@ -0,0 +1,176 @@
...
+	PhasePending Phase = "Pending"
+	PhaseRunning Phase = "Running"
+	PhaseFailed  Phase = "Failed"
+	PhaseNA      Phase = "NotApplicable"
+
+	// ConditionReady indicates the resource is ready.
+	ConditionReady ConditionType = "Ready"
+
+	// ConditionConfigured indicates valid configuration.
+	ConditionConfigured ConditionType = "Configured"
```
juliev0, line 43: so, do you think we need both of these Conditions or just one? it looks like when one is getting set true, the other one is, and when one is getting set false, the other one is

internal/controller/pipelinerollout_controller.go
juliev0: (same question for ISBServiceRollout)

pkg/apis/numaplane/v1alpha1/status_types.go
afugazzotto: good point, I'll modify it

pkg/apis/numaplane/v1alpha1/status_types.go
afugazzotto: I don't think we need both. I just was not sure which one we want to go with. Do you have a preference?

.gitignore
afugazzotto: I spent a bit of time investigating why that is being generated, but could not figure it out. So I decided to exclude it at least for now.

pkg/apis/numaplane/v1alpha1/status_types.go
afugazzotto: I'd say let's go with Configured for now.

juliev0: sounds good

juliev0 (approved):

PR #314, opened 2024-05-22
Apply NumaflowController Spec

xdevxy:
### Modifications

This PR adds the reconciliation logic for NumaflowControllerRollout.

### Verification
Tested in local cluster,

- `make start`
- `kubectl apply -f config/samples/numaplane.numaproj.io_v1alpha1_numaflowcontrollerrollout.yaml`

Verified Numaflow Controller successfully deployed to `example-namespace`.

### DIFF
--- a/internal/controller/numaflowcontrollerrollout_controller.go
+++ b/internal/controller/numaflowcontrollerrollout_controller.go
@@ -18,7 +18,11 @@ package controller
 
 import (
 	"context"
+	"fmt"
 
+	gitopsSync "github.com/argoproj/gitops-engine/pkg/sync"
+	gitopsSyncCommon "github.com/argoproj/gitops-engine/pkg/sync/common"
+	kubeUtil "github.com/argoproj/gitops-engine/pkg/utils/kube"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
@@ -26,6 +30,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	"github.com/argoproj/gitops-engine/pkg/diff"
+	"github.com/numaproj-labs/numaplane/internal/common"
+	"github.com/numaproj-labs/numaplane/internal/controller/config"
+	"github.com/numaproj-labs/numaplane/internal/metrics"
+	"github.com/numaproj-labs/numaplane/internal/sync"
 	"github.com/numaproj-labs/numaplane/internal/util/logger"
 	apiv1 "github.com/numaproj-labs/numaplane/pkg/apis/numaplane/v1alpha1"
 )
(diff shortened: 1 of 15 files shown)


internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -78,11 +131,200 @@ func (r *NumaflowControllerRolloutReconciler) Reconcile(ctx context.Context, req
...
+	if err != nil {
+		numaLogger.Error(err, "Error on comparing git sync state")
+		return gitopsSyncCommon.OperationError, err
+	}
+
+	opts := []gitopsSync.SyncOpt{
+		gitopsSync.WithLogr(*numaLogger.LogrLogger),
+		gitopsSync.WithOperationSettings(false, true, false, false),
+		gitopsSync.WithManifestValidation(true),
+		gitopsSync.WithPruneLast(false),
```
juliev0, line 275: any reason we shouldn't prune?

internal/sync/cache.go
```diff
@@ -421,13 +420,16 @@ func (c *liveStateCache) GetClusterCache() (clustercache.ClusterCache, error) {
...
+	targetObjs []*unstructured.Unstructured,
+) (map[kube.ResourceKey]*unstructured.Unstructured, error) {
 	clusterInfo, err := c.getSyncedCluster()
 	if err != nil {
-		return nil, fmt.Errorf("failed to get cluster info for %q: %w", gitsync.Spec.Destination.Cluster, err)
+		return nil, fmt.Errorf("failed to get cluster info: %w", err)
 	}
 	return clusterInfo.GetManagedLiveObjs(targetObjs, func(r *clustercache.Resource) bool {
-		return resInfo(r).GitSyncName == gitsync.Name
+		return resInfo(r).Name == name
```
juliev0, line 432: So, what does the `Name` end up getting set to? (does it uniquely identify `Kind`, `name`, and `namespace` of the resource?)

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -78,11 +131,200 @@ func (r *NumaflowControllerRolloutReconciler) Reconcile(ctx context.Context, req
...
+	phase, err := r.sync(controllerRollout, namespace, numaLogger)
+	if err != nil {
+		return err
+	}
+
+	if phase != gitopsSyncCommon.OperationSucceeded {
+		return fmt.Errorf("sync operation is not successful")
+	}
+
+	//todo: controllerRollout.Status.MarkRunning() // should already be but just in case
```
afugazzotto, line 212: Status changes have been merged to the rolloutControllers branch. You can rebase and then solve this todo.

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -78,11 +131,200 @@ func (r *NumaflowControllerRolloutReconciler) Reconcile(ctx context.Context, req
...
+			return ctrl.Result{}, err
+		}
+	}
+
+	numaLogger.Debug("reconciliation successful")
 
 	return ctrl.Result{}, nil
 }
 
+func (r *NumaflowControllerRolloutReconciler) needsUpdate(old, new *apiv1.NumaflowControllerRollout) bool {
```
afugazzotto, line 167: Now we have this function in all controllers. We should probably move it to a shared file and reuse the same on all controllers.

juliev0: right now each one checks to see if there's anything that would cause the RolloutCR to need to be updated (other than the Status). Each one really only checks for the finalizer being set/unset, although there is I suppose potential for something else to do so in the future (maybe a label or annotation?). I am okay if we want to consolidate them or just keep them separate (I imagine you want to use generics for type?)

xdevxy: yeah, I also thought about this, let's keep it separate for now, we can revisit after when reconciliation logic is more complicated.

internal/sync/cache.go
xdevxy: The `Name` is used after the matching k8s resources with the given `targetObjs` is found, to determin whether the resources are managed. And the determining logic is if the name of numaplane instance is the same.

internal/controller/numaflowcontrollerrollout_controller.go
xdevxy: we still `Prune`, just `PruneLast`, https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#prune-last

afugazzotto (approved):

internal/controller/numaflowcontrollerrollout_controller.go
juliev0: oh, okay, thanks

internal/sync/cache.go
juliev0: will it be able to distinguish between a GitSync named "pipelines" and a PipelineRollout named "pipelines", or between a PipelineRollout named "pipelines" on "namespace-a" and a PipelineRollout named "pipelines" on "namespace-b"?

juliev0: (If not, I'm okay to merge in but let's add a todo)

xdevxy: 👍  not really let me add a todo as currently we are not using gitsync

xdevxy: Done.

juliev0 (approved):

PR #319, opened 2024-05-22
Watch child resources and propagate status back - Pipelines and ISBServices

juliev0:
Adds watch to interstepbufferservice for ISB Controller and to Pipelines for PipelineController. This "magically" calls Reconcile() for the correct owning object. 

Verified with changing Pipeline object, and changing ISB object and observed that auto-healing worked.

Currently, the watch for the numaflow-controller Deployment object in the NumaflowControllerReconciler is commented out, as I wasn't able to test locally, due to an issue I haven't figured out yet.

UPDATE: also added propagation of Status of Pipeline and ISBService into the new `ConditionChildResourcesHealthy` condition and verified it seems to work for both.


afugazzotto (approved):

internal/controller/pipelinerollout_controller.go
```diff
@@ -177,8 +185,21 @@ func (r *PipelineRolloutReconciler) needsUpdate(old, new *apiv1.PipelineRollout)
...
-		// Reconcile PipelineRollouts when there's been a Generation changed (i.e. Spec change)
-		For(&apiv1.PipelineRollout{}).WithEventFilter(predicate.GenerationChangedPredicate{}).
-		Complete(r)
+
+	controller, err := runtimecontroller.New(ControllerPipelineRollout, mgr, runtimecontroller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch PipelineRollouts
```
xdevxy, line 194: Why do we need to watch PipelineRollouts specifically? I thought they are watched by default?

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -321,8 +329,35 @@ func (r *NumaflowControllerRolloutReconciler) getResourceOperations() (kubeUtil.
...
+	   		return err
+	   	}
+
+	   // Watch NumaflowControllerRollouts
+
+	   	if err := controller.Watch(source.Kind(mgr.GetCache(), &apiv1.NumaflowControllerRollout{}), &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{}); err != nil {
+	   		return err
+	   	}
+
+	   // Watch Deployments of numaflow-controller
```
xdevxy, line 351: Do we need to watch all resources deployed for numaflow-controller, like config maps?

internal/controller/pipelinerollout_controller.go
juliev0: they were being watched before in the code which this is replacing. It's just a different call

juliev0: For reference, I modeled this code on Derek's [PR](https://github.com/numaproj-labs/numaplane/pull/158/files/1148afd9fea5f766c99caa92e629527a5c15f470) which he had written by accident because he didn't realize I already had the code that was doing what he intended to do.

internal/controller/numaflowcontrollerrollout_controller.go
juliev0: Oh, yeah, you're probably right. I was just thinking about resources that we require health status for - like the numaflow-controller is Unhealthy - and we want to make sure to incorporate that. But for auto-healing, we need to be notified of it.

xdevxy (approved):

PR #320, opened 2024-05-23
Add OwnerReference to Numaflow controller

xdevxy:
### Modifications

Add owner reference when deploying Numaflow controller

### Verification

Verified if NumaflowControllerRollout CR is deleted, the underlying numaflow controller is deleted.

### DIFF
--- a/internal/controller/numaflowcontrollerrollout_controller.go
+++ b/internal/controller/numaflowcontrollerrollout_controller.go
@@ -17,8 +17,10 @@ limitations under the License.
 package controller
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 
 	"github.com/argoproj/gitops-engine/pkg/diff"
 	gitopsSync "github.com/argoproj/gitops-engine/pkg/sync"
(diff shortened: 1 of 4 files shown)


internal/util/kubernetes/kubernetes.go
```diff
@@ -133,3 +135,65 @@ func DeleteManagedObjects(ctx context.Context, client k8sClient.Client, objs map
...
+		return nil, err
+	}
+	if !found {
+		existingRefs = []interface{}{}
+	}
+
+	// Check if the owner reference already exists to avoid duplication
+	alreadyExists := ownerExists(existingRefs, ownerRef)
+
+	// Add the new owner reference if it does not exist
```
juliev0, line 184: something to think about later: are we retrying updates to objects if there's a conflict in resourceVersion? (even if we're being thread safe internally, there's potential for somebody outside of numaplane to modify them)

internal/controller/numaflowcontrollerrollout_controller.go
```diff
@@ -326,3 +350,43 @@ func (r *NumaflowControllerRolloutReconciler) SetupWithManager(mgr ctrl.Manager)
 		For(&apiv1.NumaflowControllerRollout{}).WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
 }
+
+// SplitYAMLToString splits a YAML file into strings. Returns list of yamls
+// found in the yaml. If an error occurs, returns objects that have been parsed so far too.
+func SplitYAMLToString(yamlData []byte) ([]string, error) {
```
juliev0, line 356: maybe later we can add a unit test for some of these? should we add a todo for it?

juliev0 (approved):

PR #322, opened 2024-05-24
feat: pipeline draining with image tag change

xdevxy:
### Modifications

When pipeline spec is updated with image tag change, pipeline will be paused and drained then the change will be applied and resume the pipeline to avoid data loss.

### Verification

- apply the following PipelineRollout spec,

```
apiVersion: numaplane.numaproj.io/v1alpha1
kind: PipelineRollout
metadata:
  name: my-pipeline
  namespace: example-namespace
spec:
  pipeline:
    interStepBufferServiceName: my-isbsvc
    vertices:
      - name: in
        source:
          # A self data generating source
          generator:
            rpu: 5
            duration: 5s
      - name: even-or-odd
        partitions: 2
        udf:
          container:
            # Tell the input number is even or odd, see https://github.com/numaproj/numaflow-go/tree/main/pkg/mapper/examples/even_odd
            image: quay.io/numaio/numaflow-go/map-even-odd:numaplane-demo-v1.0.0 # or use image tag map-even-odd:numaplane-demo-v1.0.1
            imagePullPolicy: Always
      - name: even-sink
        partitions: 2
        sink:
          udsink:
            container:
              # A redis sink for e2e testing, see https://github.com/numaproj/numaflow-go/tree/main/pkg/sinker/examples/redis-sink
              image: quay.io/numaio/numaflow-go/redis-sink:stable
              imagePullPolicy: Always
      - name: odd-sink
        sink:
          udsink:
            container:
              image: quay.io/numaio/numaflow-go/redis-sink:stable
              imagePullPolicy: Always
      - name: number-sink
        sink:
          udsink:
            container:
              image: quay.io/numaio/numaflow-go/redis-sink:stable
              imagePullPolicy: Always
    edges:
      - from: in
        to: even-or-odd
      - from: even-or-odd
        to: even-sink
        conditions:
          tags:
            operator: or
            values:
              - even-tag
      - from: even-or-odd
        to: odd-sink
        conditions:
          tags:
            operator: or
            values:
              - odd-tag
      - from: even-or-odd
        to: number-sink
        conditions:
          tags:
            operator: or
            values:
              - odd-tag
              - even-tag
```

- while pipeline is running, change the pipeline spec to  `duration: 1s`. This pipeline spec change will be applied immediately.
- Change the pipeline spec to have `even-or-odd` vertex use a a different image `image: quay.io/numaio/numaflow-go/map-even-odd:numaplane-demo-v1.0.0 # or use image tag map-even-odd:numaplane-demo-v1.0.1` 
- This pipeline spec will trigger a pausing and after data is drained the pipeline spec is updated and resumed to avoid data loss.

### DIFF
--- a/internal/controller/pipelinerollout_controller.go
+++ b/internal/controller/pipelinerollout_controller.go
@@ -18,9 +18,11 @@ package controller
 
 import (
 	"context"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -98,7 +100,7 @@ func (r *PipelineRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	pipelineRollout.Status.InitConditions()
 
-	err := r.reconcile(ctx, pipelineRollout)
+	enqueue, err := r.reconcile(ctx, pipelineRollout)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
@@ -122,13 +124,19 @@ func (r *PipelineRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 	}
 
+	if enqueue {
+		return ctrl.Result{Requeue: true, RequeueAfter: 30 * time.Second}, nil
+	}
 	numaLogger.Debug("reconciliation successful")
 
 	return ctrl.Result{}, nil
 }
 
 // reconcile does the real logic
-func (r *PipelineRolloutReconciler) reconcile(ctx context.Context, pipelineRollout *apiv1.PipelineRollout) error {
+func (r *PipelineRolloutReconciler) reconcile(
+	ctx context.Context,
+	pipelineRollout *apiv1.PipelineRollout,
+) (bool, error) {
 	numaLogger := logger.FromContext(ctx)
 
 	// is PipelineRollout being deleted? need to remove the finalizer so it can
@@ -138,7 +146,7 @@ func (r *PipelineRolloutReconciler) reconcile(ctx context.Context, pipelineRollo
 		if controllerutil.ContainsFinalizer(pipelineRollout, finalizerName) {
 			controllerutil.RemoveFinalizer(pipelineRollout, finalizerName)
 		}
-		return nil
+		return false, nil
 	}
 
 	// add Finalizer so we can ensure that we take appropriate action when CRD is deleted
@@ -146,6 +154,8 @@ func (r *PipelineRolloutReconciler) reconcile(ctx context.Context, pipelineRollo
 		controllerutil.AddFinalizer(pipelineRollout, finalizerName)
 	}
 
+	// if not, check the change if it is image change, if yes, pause
+	// if not, just update
 	// apply Pipeline
 	// todo: store hash of spec in annotation; use to compare to determine if anything needs to be updated
 	obj := kubernetes.GenericObject{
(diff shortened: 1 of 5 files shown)


internal/controller/pipelinerollout_controller.go
```diff
@@ -161,24 +171,80 @@ func (r *PipelineRolloutReconciler) reconcile(ctx context.Context, pipelineRollo
...
-	}
+		// create object as it doesn't exist
+		if apierrors.IsNotFound(err) {
+			err = kubernetes.CreateCR(ctx, r.restConfig, &obj, "pipelines")
+			if err != nil {
+				return false, err
+			}
+		}
+	} else {
+		// If the pipeline already exist, first check if the pipeline status
```
juliev0, line 185:
```suggestion
		// If the pipeline already exists, first check if the pipeline status
```

internal/controller/pipelinerollout_controller.go
```diff
@@ -122,13 +124,19 @@ func (r *PipelineRolloutReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 
+	if enqueue {
+		return ctrl.Result{Requeue: true, RequeueAfter: 30 * time.Second}, nil
+	}
 	numaLogger.Debug("reconciliation successful")
 
 	return ctrl.Result{}, nil
 }
 
 // reconcile does the real logic
```
juliev0, line 135: can you clarify what the bool being returned represents in the comment?

internal/controller/pipelinerollout_controller.go
```diff
@@ -161,24 +171,80 @@ func (r *PipelineRolloutReconciler) reconcile(ctx context.Context, pipelineRollo
...
+			return false, nil
+		}
+
+		// If pipeline status is not above, detect if pausing is required.
+		shouldPause, err := needsPausing(pipeline, &obj)
+		if err != nil {
+			return false, err
+		}
+		if shouldPause {
+			// Use the existing spec, then pausing and re-enqueue
```
juliev0, line 223:
```suggestion
			// Use the existing spec, then pause and re-enqueue
```

internal/controller/pipelinerollout_controller.go
```diff
@@ -236,3 +302,125 @@ func (r *PipelineRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	return nil
 }
+
+func setPipelineDesiredStatus(obj *kubernetes.GenericObject, status string) (*kubernetes.GenericObject, error) {
+	unstruc, err := kubernetes.ObjectToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	err = unstructured.SetNestedField(unstruc.Object, status, "spec", "lifecycle", "desiredPhase")
```
juliev0, line 312: oh, nice, I didn't know about this method.

juliev0: I wonder if we should ultimately have a `GenericObject.SetNestedField()` method that this could call

internal/controller/pipelinerollout_controller.go
```diff
@@ -236,3 +302,125 @@ func (r *PipelineRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	numaLogger := logger.FromContext(ctx)
+	pipelineStatus, err := kubernetes.ParseStatus(pipeline)
+	if err != nil {
+		numaLogger.Errorf(err, "failed to parse Pipeline Status from pipeline CR: %+v, %v", pipeline, err)
+		return false
+	}
+
+	numaLogger.Debugf("pipeline status: %+v", pipelineStatus)
+
+	return numaflowv1.PipelinePhase(pipelineStatus.Phase) == numaflowv1.PipelinePhasePaused
```
juliev0, line 347: could potentially take the two methods above and strip out the shared logic into a common function

internal/controller/pipelinerollout_controller.go
```diff
@@ -236,3 +302,125 @@ func (r *PipelineRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		numaLogger.Errorf(err, "failed to get Pipeline: %v", err)
+		return err
+	}
+
+	processPipelineStatus(ctx, pipeline, pipelineRollout)
+	return nil
+}
+
+// TODO: detect engine
+func needsPausing(obj *kubernetes.GenericObject, newObj *kubernetes.GenericObject) (bool, error) {
```
juliev0, line 376: since this is an optimization, I wonder if when we merge this in, we should just have this return true for now? (like first make sure we can always just pause reliably before trying to get fancy)

xdevxy: yes :)

